PR #7, opened 2019-04-22
Support merging swagger definitions

puneetbehl:


puneetbehl:
@graemerocher Tests are failing because some changes are missing from 1.1.x so I think once you merge these changes the tests should pass.

PR #9, opened 2019-04-26
Default the Operation Summary To Be The Same Value As The Description

recursivecodes:
OpenAPI uses the summary in the API UI to put a friendly description to the endpoint.  This can be accomplished via manually adding an `@Operation(summary = "foo")` annotation, but if we have a description we should default it to that.

image


puneetbehl:
@recursivecodes Thanks for the PR the changes will be merged onto 1.1.x in PR #16

PR #31, opened 2019-07-24
Correct security requirement object definition

dstepanov:
According to https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#securityRequirementObject and Swagger UI

Correct format should be:
```
custom_name:
        - scope1
        - scope2
```
not
```
name:
   - custom_name
scopes:
   - scope1
   - scope2 
```


graemerocher:
Thanks!

lordaronwd:
Hi guys, thanks for fixing this!
Any idea when the fix is going to be released?

PR #47, opened 2019-08-12
Do not generate request body if all parameters are ignored

### LINKED ISSUE #44: Request body generated in Swagger YAML for POST/PUT with no body
This is to reopen https://github.com/micronaut-projects/micronaut-core/issues/1200

I found the related code here https://github.com/micronaut-projects/micronaut-openapi/blob/23409f4537aa64f07f0558a6ca9c92e797536127/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiControllerVisitor.java#L435

If all the body parameters are skipped, e.g. all of them are hidden parameters, then the body will be empty. In this case, I think body should not be required as in
https://github.com/micronaut-projects/micronaut-openapi/blob/23409f4537aa64f07f0558a6ca9c92e797536127/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiControllerVisitor.java#L459

Consequences (not related to this repo): when I use codegen to generate client code from the generated swagger.yml, it requires me to pass in an empty object although it shouldn't

HoaBo:
fix #44

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiControllerVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiControllerVisitor.java
@@ -417,9 +417,17 @@ public class OpenApiControllerVisitor extends AbstractOpenApiVisitor implements 
             }
 
             if (HttpMethod.requiresRequestBody(httpMethod) && swaggerOperation.getRequestBody() == null) {
-                List<ParameterElement> bodyParameters = Arrays.stream(element.getParameters()).filter(p -> !pathVariables.containsKey(p.getName()) && !p.isAnnotationPresent(Bindable.class)).collect(Collectors.toList());
-                if (!bodyParameters.isEmpty()) {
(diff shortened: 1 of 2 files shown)


graemerocher:
Thanks for the PR!

HoaBo:
@graemerocher I'd like use this fix in the snapshot version. However when I try to use https://oss.sonatype.org/content/repositories/snapshots/io/micronaut/micronaut-bom/1.2.1.BUILD-SNAPSHOT/micronaut-bom-1.2.1.BUILD-20190812.213052-3.pom, <micronaut.openapi.version> is still at 1.2.0
Could you please let me know how to bump that to snapshot version as well? Or how can I use an out-of-band version for micronaut-openapi?

HoaBo:
Nvm, I found the way. Just note it here in case anyone needs it in the future

```
repositories {
    maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
}

dependencies {
    annotationProcessor "io.micronaut.configuration:micronaut-openapi:1.2.1.BUILD-20190813.063334-3"
    implementation "io.swagger.core.v3:swagger-annotations"
}
```

PR #48, opened 2019-08-13
Observable should return an array of elements

HoaBo:
Micronaut says that it supports io.reactivex.Observable
In that case, the response will be an array of elements


HoaBo:
Closing this as it conflicts with master after another PR was merged

PR #51, opened 2019-08-14
Return value should be allowed to be multiline

### LINKED ISSUE #45: OpenAPI yaml is incorrect if Javadoc @return tag contains a multi-line string
Context: Micronaut 1.2.0 Java project generated with command

```console
mn create-app --features=swagger-java micronaut-openapi-parse-javadoc
```

If the @return tag of the Javadoc of a controller contains a multi-line string the generated OpenAPI yaml is incorrect.

Example of such controller:

```java
  /**
   * This is a
   * multiline description
   * of the API
   * @param testParam This is a
   *                  multiline description
   *                  of the parameter
   * @return This is a
   *     multiline description
   *     of the return value
   */
  @Get("/{testParam}")
  public Map<String, String> index(String testParam) {
  [...]
```

Expected OpenAPI yaml:

```yml
paths:
  /{testParam}:
    get:
      summary: |-
        This is a
                       multiline description
                       of the API
      description: |-
        This is a
                       multiline description
                       of the API
      operationId: index
      parameters:
      - name: testParam
        in: path
        description: |-
          This is a
                            multiline description
                            of the parameter
        required: true
        schema:
          type: string
      responses:
        default:
          description: |-
            This is a
                             multiline description
                             of the return value
          content:
            application/json:
              schema:
                type: object
                additionalProperties:
                  type: string
```

Actual OpenAPI yaml:

 ```yml
paths:
  /{testParam}:
    get:
      summary: "This is a\n multiline description\n of the API\n \n     multiline\
        \ description\n     of the return value"
      description: "This is a\n multiline description\n of the API\n \n     multiline\
        \ description\n     of the return value"
      operationId: index
      parameters:
      - name: testParam
        in: path
        description: |-
          This is a
                            multiline description
                            of the parameter
        required: true
        schema:
          type: string
      responses:
        default:
          description: This is a
          content:
            application/json:
              schema:
                type: object
                additionalProperties:
                  type: string
```

As you can see multiline strings in @param tags are correctly handled, but the presence of multiline string in the @return tag causes the description of the response to only contain the first line, and the following lines are instead appended to the summary and description of the operation.

I have set up an example repository that can be used to reproduce the problem: https://github.com/Rosetjau/micronaut-openapi-parse-javadoc

JacobAae:
Previously it switched to TEXT state on newline which would add remainder to description.
Fixes #45

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/javadoc/JavadocParser.java
+++ b/openapi/src/main/java/io/micronaut/openapi/javadoc/JavadocParser.java
@@ -63,17 +63,13 @@ public class JavadocParser {
             for (char c : chars) {
                 switch (state) {
                     case RETURN_DESC:
-                        if (c == '\n') {
-                            state = TEXT;
-                            previousState = TEXT;
-                            javadocDescription.setReturnDescription(currentDescription.toString());
-                            currentDescription = description;
-                        }
+                        javadocDescription.setReturnDescription(currentDescription.toString().trim());
+                        // Notice pass through in case
                     case PARAM_DESC:
                         if (state == PARAM_DESC) {
                             javadocDescription.getParameters().put(currentParam.toString(), currentDescription.toString().trim());
                         }
-
+                        // Notice pass through in case
                     case TEXT:
                         if (c == '{' || c == '@') {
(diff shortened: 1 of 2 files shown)


graemerocher:
Thanks for the PR!

Issue #46, opened 2019-08-11
Generated doc using Operation and ApiResponse are missing summary, responseCode, description and mediaType warning


geemang2000:
Version: 1.2
Project Language: groovy

1. Operation summary and description are not generated in the yml (see screenshots)

1. Adding mediaType to Content annotation generates a warning in the build process
```
WARNING: Error reading Swagger Operation for element [io.micronaut.ast.groovy.visitor.GroovyMethodElement@ac1683d]: Cannot construct instance of `io.swagger.v3.oas.models.media.MediaType` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('application/json')
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: io.swagger.v3.oas.models.responses.ApiResponse["content"]->io.swagger.v3.oas.models.media.Content["mediaType"]) (through reference chain: io.swagger.v3.oas.models.Operation["responses"])

       @Operation(summary = "returns a list of applications",
       ^
Note: Writing OpenAPI YAML to destination: file:/Users/martingee/dev/iamportal/server/build/classes/groovy/main/META-INF/swagger/icsynergy-idaas-gateway-1.0.yml
```
2. Loading the generated yml in (editor.swagger.io) has an error every endpoint. _Should the description should be auto-generated?_.
```
Structural error at paths./api/v1/app.get.responses.default
should have required property 'description'
missingProperty: description
Jump to line 32
```
3.  NOTE: in my example (below) the description and responseCode are not generated in the yml. See screenshots.  Seems to be ignoring. 

```
    @Operation(summary = "returns a list of applications",
            description = "Retrieve all available applications",
            responses = [
                    @ApiResponse(
                            responseCode = "200",
                            description = "The Application List",
                            content = @Content(mediaType = MediaType.APPLICATION_JSON,
                            array = @ArraySchema(schema = @Schema(implementation = Application.class))))
            ]
    )
    @Get
    List<Application> getAllApps() {
        log.info "getAllApps"
        appService.getAllApps()
    }
```

![Screen Shot 2019-08-11 at 7 42 52 AM](https://user-images.githubusercontent.com/392233/62833946-e722f200-bc0b-11e9-86ce-f286778b7060.png)

![Screen Shot 2019-08-11 at 7 47 21 AM](https://user-images.githubusercontent.com/392233/62834079-03735e80-bc0d-11e9-90ce-e48f3490d2bd.png)

[...]

PR #54, opened 2019-08-15
Default response description is required

HoaBo:
When the generated yml file is validated, a bunch of errors show up because the default response description is missing. This issue is also mentioned in #46 

I'm not sure what is a sensible default response description, so I choose `openrationId + " default response"`

This PR also add an assert for the test case when requestBody is null


graemerocher:
Thanks!

PR #64, opened 2019-08-29
Support for @ExampleObject and @RequestBody

krzykrucz:
Added support for @ExampleObject and @RequestBody annotations, as well as jsons nested as a String (to be parsed by yaml)


graemerocher:
Thanks for the contribution!

PR #66, opened 2019-08-29
Add option to generate views for generated OpenApi yaml.

croudet:
Currently, views for swagger-ui, redoc and rapidoc are supported.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/view/OpenApiViewConfig.java
+++ b/openapi/src/main/java/io/micronaut/openapi/view/OpenApiViewConfig.java
@@ -1,0 +1,503 @@
+/*
+ * Copyright 2017-2019 original authors
+ *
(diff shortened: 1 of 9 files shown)


graemerocher:
Thanks this looks like a useful PR. I wonder if the dependency on handlebars is necessary though.

croudet:
The dependency is only use at compile time. But the templates are simple enough so that I can get rid of handlerbars if you want.

graemerocher:
That would be good I think. The fewer dependencies the better, especially in annotation processor scope

croudet:
Handlebars dependency has been removed.

graemerocher:
Thanks this is a great PR, one thing do the docs only mention how to setup Gradle, do you know what the equivalent Maven config would be and whether that can be added to the documentation?

croudet:
I have added the maven configuration.

graemerocher:
Thanks for the contribution!

PR #65, opened 2019-08-29
Fixed required properties for a Schema

krzykrucz:
section `required` wasn't being added to Schema object. It's a fix for that

PR #74, opened 2019-09-19
Fix duplicate required properties

croudet:
https://github.com/micronaut-projects/micronaut-openapi/pull/65
introduces support for 'requiredProperties' attribute, but it can lead to have duplicate entries
in the generated yaml.


graemerocher:
Thanks!

PR #75, opened 2019-09-22
Indicate that the swagger support does not support Groovy.

pmcneil:
Currently there is no indication that Swagger support does not support Groovy unless you happen to try the cli:
mn create-app my-openapi-app --features swagger-java --lang groovy
If you are trying to add swagger support to an existing groovy project it just doesn't work.

### DIFF
--- a/src/main/docs/guide/introduction.adoc
+++ b/src/main/docs/guide/introduction.adoc
@@ -1,3 +1,5 @@
 Micronaut includes experimental support for producing OpenAPI (Swagger) YAML at compilation time. Micronaut will at compile time produce a Swagger 3.x compliant YAML file just based on the regular Micronaut annotations and the javadoc comments within your code.
 
+This does *not* currently support Groovy.
+
 You can customize the generated Swagger using the standard https://github.com/swagger-api/swagger-core/wiki/Swagger-3.X---Annotations[Swagger Annotations].


graemerocher:
It does not? Wow major bug then, it should support Groovy

graemerocher:
How did you reach the assertion that it doesn't work for Groovy?

I tried the following steps:

```
$ sdk use micronaut 1.2.2
$ mn create-app groovy-openapi --lang groovy --features swagger-groovy
$ cd groovy-openapi
$ mn create-controller foo
$ ./gradlew assemble
```

And the  Swagger YAML was generated to `build/classes/groovy/main/META-INF/swagger/groovy-openapi-0.0.yml`

pmcneil:
on Linux:
[pmcneil@big-nerd1 micronaut]$ mn create-app my-openapi-app --features swagger-java --lang groovy
| Generating Java project...
| Warning The following features are incompatible with other feature selections and have been removed from the project
| groovy
| Application created at /home/pmcneil/devel/micronaut/my-openapi-app

ilopmar:
@pmcneil you are using the wrong feature. It needs to be `swagger-groovy` instead of `swagger-java`

pmcneil:
OK. Cool. so the documentation needs to point out that there is swagger-groovy exists?? I'll double check, but I couldn't see that anywhere and frankly just missed it in Graemes comment.

PR #86, opened 2019-11-14
allOf is not appearing properly in generated schema

croudet:
Given the following classes with an `allOf`
```java

@Schema(description = "A")
public class A {
    @Schema(description = "a")
    @Nullable
    private String a;

    public String getA() {
        return a;
    }

    public void setA(String a) {
        this.a = a;
    }

}

@Schema(description = "B", allOf = A.class)
public class B extends A {
    @Schema(description = "b")
    @Nullable
    private String b;

    public String getB() {
        return b;
    }

    public void setB(String b) {
        this.b = b;
    }

}
```
Micronaut generates
```
components:
  schemas:
    A:
      type: object
      properties:
        a:
          type: string
          description: a
          nullable: true
      description: A
    B:
      type: object
      properties:
        b:
          type: string
          description: b
          nullable: true
      description: B
      allOf:
      - $ref: '#/components/schemas/A'
```
While what is expected is:
```
components:
  schemas:
    A:
      type: object
      properties:
        a:
          type: string
          description: a
          nullable: true
      description: A
    B:
      type: object
      description: B
      allOf:
      - $ref: '#/components/schemas/A'
      - type: object
        properties:
          b:
            type: string
            description: b
            nullable: true
```
Reference: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -699,6 +699,18 @@ abstract class AbstractOpenApiVisitor  {
         }
     }
 
+    private void checkAllOf(ComposedSchema composedSchema) {
+        if (composedSchema != null && composedSchema.getAllOf() != null && !composedSchema.getAllOf().isEmpty()) {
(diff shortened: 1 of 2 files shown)


croudet:
Not ready

PR #97, opened 2019-11-22
fix maven compiler argument to generate openapi views.

croudet:


witek1902:
I had the same problem, fix from PR works.

PR #101, opened 2019-11-28
maven compile args requires "-J" prefix

pjungermann:
see "-Joption" at https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javac.html#options

otherwise, it results in an "invalid flag" error


pjungermann:
I just realized that there is already PR #97 which will fix this, too.

PR #106, opened 2019-12-04
Context path aware openapi views.

croudet:
Some logic added for the open views to work in case a server context path is defined

You need to pass a parameter to the view with the defined context path:

http://localhost:8080/context/path/swagger-ui?contextPath=%2Fcontext%2Fpath.


graemerocher:
Thanks for the contribution!

Issue #24, opened 2019-05-28
Swagger file generation does not obey snake_case setting for jackson


lenalebt:
I have a project where I configured jackson to use `snake_case` in JSON:

From `application.yml`:
```
jackson:
  property-naming-strategy: SNAKE_CASE
```

Schemas generated for my Kotlin data classes are generated with `camelCase`:
```
data class Foo(val hasChildren: Boolean)
```
leads to
```
components:
  schemas:
    Foo:
      type: object
      properties:
        hasChildren:
          type: boolean
```

`hasChildren` should be named `has_children`.

Are there any known workarounds?

graemerocher:
Not currently, It would likely have to be a configuration option. Currently it defaults to use `getName()` of the `PropertyElement` on this line:

https://github.com/micronaut-projects/micronaut-openapi/blob/master/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L418

It is a little complex because currently we do parse the `application.yml` configuration at Compile time so the best bet would be some other annotation to the control the behaviour.

sdelamo:
@lenalebt  The workaround we are using is using the @Schema annotation in our POJOs and specify the name attribute:

```
import io.swagger.v3.oas.annotations.media.Schema;

@Introspected
public class RefreshRequest {

    @Schema(description = "Grant Type", example = "refresh_token", required = true, name = "grant_type")
    @NotBlank
    @Pattern(regexp = "refresh_token")
    @JsonProperty("grant_type")
    private String grantType;
```

lenalebt:
Hi @sdelamo thanks for that tip, I'll give it a try :)!

picaso:
Hi! 
Any update on this? Trying to use Micronaut in our org, and this was flagged as an issue, the work-around does work though, just something else to maintain.

mkfuri:
Hi! I've the same problem. Any update on this?

gkarnik05:
In Asp.Net core, the following 2 lines helped to generate Swagger UI schema in snake case:
services.AddMvc()
             .AddNewtonsoftJson(x =>
             {
                 x.SerializerSettings.ContractResolver = new DefaultContractResolver
                 {
                     NamingStrategy = new SnakeCaseNamingStrategy()
                 };
             });
services.AddSwaggerGenNewtonsoftSupport();
Reference: https://github.com/domaindrivendev/Swashbuckle.AspNetCore

PR #107, opened 2019-12-09
Specify Property naming strategy a compile time.

croudet:
Should fix https://github.com/micronaut-projects/micronaut-openapi/issues/24

based on the following:
https://gist.github.com/silas/3b41e6ed4031c11f919767a74829ffea
https://github.com/swagger-api/swagger-core/blob/26d72bf5fd2d54e5e514ddcf3861d125d2f95b42/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/override/SnakeCaseConverterTest.java

Uses a system property to specify the schema naming strategy property.


graemerocher:
Wow, thanks for the all the great contributions

Issue #110, opened 2019-12-13
generated security declaration in yaml has wrong format


dlamoris:
Hi, I have a controller method annotated like this:
```
 @Operation(operationId = "doStuff", security = @SecurityRequirement(name = "BasicAuth", scopes={}))
```

and in the generated yaml, it has this
```
security:
  - name:
    - BasicAuth
```

but according to the openapi doc (https://swagger.io/docs/specification/authentication/basic-authentication/), it should be
```
security:
  - BasicAuth: []
```

croudet:
I think a current work around is to use `SecuritryRequirement` outside `Operation`:
```
@SecurityRequirement(name = "BasicAuth", scopes={})
@Operation(operationId = "doStuff")
```

dlamoris:
thanks, that works!

PR #112, opened 2019-12-13
Security requirement

croudet:
Fix processing of SecurityRequirement when using in Operation.
Should fix https://github.com/micronaut-projects/micronaut-openapi/issues/110


puneetbehl:
@croudet It seems like a bug fix, could you please send the PR against branch 1.2.x

croudet:
Will redo a PR against 1.2.x.

Issue #130, opened 2020-02-06
POST service: parameter duplicated in Header and body 


outersky:
## Case 1:

```kotlin
    @Post("/create2")
    fun create2(@Header("X-Session-Id") sessionId: String, phone: String, name: String): String {
        return """{"sessionId":"$sessionId", "phone":"$phone", "name":"$name"}"""
    }
```

create2: sessionId will appear twice in both header and body:

```yml
/create2:
    post:
      operationId: create2
      parameters:
      - name: X-Session-Id
        in: header
        required: true
        schema:
          type: string
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                sessionId:
                  type: string
                  nullable: false
                phone:
                  type: string
                  nullable: false
                name:
                  type: string
                  nullable: false
        required: true
      responses:
        default:
          content:
            application/json:
              schema:
                type: string
```

## Case 2

```kotlin
    @Get("/find1")
    fun find2(name: String): String {
        return """{"name":"$name"}"""
    }
```
find1: name is ignored in swagger file, but it's a query parameter by default in micronaut.

```yml
/find1:
    get:
      operationId: find2
      parameters: []
      responses:
        default:
          content:
            application/json:
              schema:
                type: string
```

I created a [demo project here](https://github.com/outersky/micronaut-openapi-bug-yml.git)

croudet:
I am not sure I understand the problem with case 2. `name` does not appear in your uri.

outersky:
@croudet  For case 2, "name" should be included in parameters:[] 

    curl http://localhost:8080/find1?name=outersky

croudet:
For case #2, I think you still need to declare it in the uri, for instance for a `PathVariable`:
`@Get(/find1/{name})`
or `QueryValue`:
`@Get(/find1{?name}`

outersky:
I know that will work , just feel like a little bit noisy and result in a lot of work . 
As in case 1, both `phone` and `name` need not be annotated, and it will generate the right openapi file.
so I think it might be better if we can recognize the default behaviour in micronaut .

croudet:
The two merged pull requests should fix both issues.

outersky:
thanks , but it's still not ok with case #2

PR #132, opened 2020-02-06
Issue # 130 - Case 2

croudet:
A parameter with no annotation is considered as a PathVariable.

https://github.com/micronaut-projects/micronaut-openapi/issues/130


graemerocher:
Maybe needs a test

croudet:
Tests added.

graemerocher:
Thanks 👍

outersky:
@croudet Default as a PathVariable but not a  QueryValue ?

With my [test project](https://github.com/outersky/micronaut-openapi-bug-yml.git)

the correct request is :
```
curl http://localhost:8080/find1?name=outersky
```

but now [swagger-ui](http://localhost:8080/swagger)  sends a request as :

```
curl -X GET "http://localhost:8080/find1" -H "accept: application/json"
```

which will lead to an error:

```json
{
  "message": "Required argument [String name] not specified",
  "path": "/name",
  "_links": {
    "self": {
      "href": "/find1",
      "templated": false
    }
  }
}
```

PR #139, opened 2020-02-11
Add support for rapipdf

croudet:
rapipdf allows to generate a PDF from a spec file (in the browser).

Adds a button to rapidoc view when enabled.


graemerocher:
Nice work again!

PR #142, opened 2020-02-12
Minor fixes:

croudet:
- documentation
- remove spec-url from valid options
- add style (css) as valid option.
- add RapiPdf to all renderers
- fix some warnings
- Handle @ArraySchema with arraySchema field.


graemerocher:
Thanks!

Issue #144, opened 2020-02-17
Micronaut for Spring Boot ignores path on @RestController


mikaelvik:
I've created an [example project](https://gitlab.com/mvik-micronaut/mvik-openapi/-/tree/master/mvik-openapi-microboot) using Micronaut to generate OpenAPI yaml for a Spring Boot application. 

Micronaut seems to ignore the path specified on `@RestController` annotations when generating the yaml file. 

The only Swagger annotation I've used is `@OpenAPIDefinition` on the Micronaut application class (and I'd like to keep it that way). 

Controller: 

```java
@RestController
@RequestMapping("/api/goodbyes") // edited
public class GoodbyeRestController {

    @GetMapping("/{name}")
    Map<String, String> goodbye(@PathVariable String name) {
        return Collections.singletonMap("goodbye", name);
    }
}
```

Excerpt from openapi.gradle (applied by build.gradle): 

```groovy
sourceSets {
    openapi {
        compileClasspath += sourceSets.main.compileClasspath
        java {
            srcDirs = [
                'src/main/java',
                'src/openapi/java'
            ]
        }
    }
}
```

Application class:

```java
@OpenAPIDefinition(info = @Info(title = "mvik-openapi-microboot", version = "0.1"))
public class MicronautApplication {

    public static void main(String[] args) {
        Micronaut.run(MicronautApplication.class);
    }
}
```
Output:

```yaml
openapi: 3.0.1
info:
  title: mvik-openapi-microboot
  version: "0.1"
paths:
  /{name}:
    get:
      operationId: goodbye
      parameters:
      - name: name
        in: path
        required: true
        schema:
          type: string
      responses:
        default:
          description: goodbye default response
          content:
            application/json:
              schema:
                type: object
                additionalProperties:
                  type: string
```

graemerocher:
According to https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RestController.html

The value of `@RestController` represents the bean name in Spring and not the URI. I believe to get the behaviour you want have to use:

```java
@RestController
@RequestMapping("/api/goodbyes")
```

mikaelvik:
Sorry for the terrible example code, the problem is also there when using `@RequestMapping`. I've used the same approach on an existing work project where all the `@RequestMapping` paths were ignored. 

I've changed the example code, and the path is still being ignored.

mikaelvik:
[...]

PR #150, opened 2020-02-17
#144 - Use the value of the UriMapping annotation on the element's de…

mikaelvik:
…claring type instead of Controller.

No new tests have been added due to this problem being discovered when using micronaut-spring. All existing tests still pass.

Resolving UriMapping directly on the element returned the same value as HttpMethodMapping, but using `element.getDeclaringType().getValue(UriMapping) seemed to work.


graemerocher:
Thanks for the contribution!

Issue #129, opened 2020-01-30
Description from Javadoc @param not generated when first line is empty


meredrica:
Given the following Javadoc:
```
/**
* Check if the given user has access.
*
* @param username The username to check.
* @param password The password to check.
* @return 'allow' or 'deny'.
*/
```
The generated API doc is correct. However, with this:

```
/**
* Check if the given user has access to RabbitMQ.
*
* @param username
*                   The username to check.
* @param password
*                   The password to check.
* @return 'allow' or 'deny' and user tags.
*/
```

The generated yaml file lacks the description.

Expected:

```
      requestBody:
        content:
          application/x-www-form-urlencoded:
            schema:
              type: object
              properties:
                username:
                  type: string
                  description: "The username to check."
                  nullable: false
                password:
                  type: string
                  description: The password to check.
                  nullable: false
```
Actual:

```
      requestBody:
        content:
          application/x-www-form-urlencoded:
            schema:
              type: object
              properties:
                username:
                  type: string
                  nullable: false
                password:
                  type: string
                  nullable: false
```

PR #152, opened 2020-02-17
Do not append '\n' for param name.

croudet:
May fix https://github.com/micronaut-projects/micronaut-openapi/issues/129


graemerocher:
Thanks for all the PRs @croudet 

Do you think we are really to release 1.4.0?

croudet:
Yes I think we are ready for a new release.

Thanks

PR #170, opened 2020-03-10
Prepend paths with context server path and expand properties

croudet:
Adds the ablility to prepend paths with a user defined server context path,
and to expand user defined properties in open api text fields.


graemerocher:
Thanks!

PR #174, opened 2020-03-24
Fix context path cookie http filter in documentation and 

croudet:
add Redoc's expandSingleSchemaField option

 new option in https://github.com/Redocly/redoc/blob/master/CHANGELOG.md#200-rc24-2020-03-17


graemerocher:
Thanks!

PR #179, opened 2020-03-27
Use path relative url for spec file

croudet:


croudet:
Well I think the way it is currently is correct. So I am closing this PR.

Issue #193, opened 2020-04-09
Incorrect schema reference when extending a generic controller


gabordicso:
The generated OpenAPI definition .yml contains an incorrect schema reference when extending a generic controller, unless the endpoint method is explicitly overridden in the child class. See the example below. The method uri is specified in the base class method's Get annotation, but the Controller path is specified in the Controller annotation of the child class, so for the endpoint path constructed from the two, I would expect the generated API definition to reference the generic type of the child class, not the base class. Example:

Micronaut version: 1.3.1

TestBaseObject:

`open class TestBaseObject (var str: String)`

TestChildObject:

`class TestChildObject (str: String, var id: Long? = null) : TestBaseObject(str)`

TestBaseService (this is the generic controller to be extended):

```
open class TestBaseService<T: TestBaseObject>
constructor() {
    lateinit var obj: T

    @Get(uri = "testMethodOnlyInBase")
    fun testMethodOnlyInBase(): T {
        return obj
    }
}
```

TestChildService (this is the actual controller to generate API definition for):

```
@Controller("/api/v1/test")
class TestChildService: TestBaseService<TestChildObject>() {
}
```

Expected .yml output:

```
  /api/v1/test/testMethodOnlyInBase:
    get:
      operationId: testMethodOnlyInBase
      parameters: []
      responses:
        default:
          description: testMethodOnlyInBase default response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TestChildObject'
```

Actual .yml (difference in last line):

```
  /api/v1/test/testMethodOnlyInBase:
    get:
      operationId: testMethodOnlyInBase
      parameters: []
      responses:
        default:
          description: testMethodOnlyInBase default response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TestBaseObject'
```

Link to complete example with additional use cases covered: [https://github.com/gabordicso/public/tree/master/micronaut/openapi](https://github.com/gabordicso/public/tree/master/micronaut/openapi)

istvanszoboszlai:
Hello, is there something blocking with fixing this issue? Do you guys need help?

graemerocher:
@croudet is https://github.com/micronaut-projects/micronaut-core/pull/3069 related?

croudet:
In some way it is. As some classes and mehtods are visited when they don't have the required annotation or when overriden methods are also processed.

The other part of the problem I thin is this call:
[...]

PR #199, opened 2020-04-14
Issue #193 - Inherited methods are not processed

croudet:
`TypeElementVisitor.visitMethod(MethodElement element, VisitorContext context)` is not called for inherited methods.

Use resolved generic types when needed.

https://github.com/micronaut-projects/micronaut-openapi/issues/193


croudet:
not yet ready

Issue #204, opened 2020-04-19
Bad option name error with 1.4.4


graemerocher:
Error:

```> Task :starter-api:compileJava FAILED
error: Bad option name 'micronaut.openapi.server.context-path' provided by processor 'org.gradle.api.internal.tasks.compile.processing.TimeTrackingProcessor'
Note: Generating OpenAPI Documentation
Note: Writing OpenAPI YAML to destination: /Users/graemerocher/dev/micronaut/starter/starter-api/build/classes/java/main/META-INF/swagger/micronaut-starter-1.0.yml
Note: Writing OpenAPI View to destination: /Users/graemerocher/dev/micronaut/starter/starter-api/build/classes/java/main/META-INF/swagger/views/redoc/index.html
Note: Writing OpenAPI View to destination: /Users/graemerocher/dev/micronaut/starter/starter-api/build/classes/java/main/META-INF/swagger/views/rapidoc/index.html
Note: Writing OpenAPI View to destination: /Users/graemerocher/dev/micronaut/starter/starter-api/build/classes/java/main/META-INF/swagger/views/swagger-ui/index.html
Note: Creating bean classes for 5 type elements
error: Bad option name 'micronaut.openapi.server.context-path' provided by processor 'org.gradle.api.internal.tasks.compile.processing.SupportedOptionsCollectingProcessor'
2 errors
```

Steps to reproduce:

```
git clone git@github.com:micronaut-projects/micronaut-starter.git
cd micronaut-starter
```

Edit version in `starter-api/build.gradle` and set to 1.4.4

```
./gradlew starter-api:cJ
```

PR #205, opened 2020-04-19
Issue #204 - Bad option name 'micronaut.openapi.server.context-path' …

croudet:
…provided by processor

Looks like '-' is not allowed in processor option name.
https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Processor.html#getSupportedOptions()

renamed `micronaut.openapi.server.context-path` to `micronaut.openapi.server.context.path`


graemerocher:
Thanks!

Issue #220, opened 2020-05-05
Annotation processor fails if method annotation contains multiple uris and not single uri


sumitaneja:
Controller method annotated like below gives compile error 
**Path variable name: 'pv' not found in path.**

**@Post(uris = {
            "/abc/xyz/{pv}/sth",
            "/abc/efg/{pv}/sth"
    })**

below works
**@Post(uri = {
            "/abc/xyz/{pv}/sth")**

PR #222, opened 2020-05-05
Issue #220 - Check if http method has multiple uris.

croudet:


graemerocher:
Thanks for all the PRs!

sumitaneja:
@croudet @graemerocher Thanks Validated fix on the snapshot. 
 looking for this fix via release channel.

PR #223, opened 2020-05-06
Upgrade to Micronaut 2.x

alvarosanchez:
See https://github.com/micronaut-projects/micronaut-core/issues/3024


croudet:
Is `getPropertyPlaceholderResolver` really needed?

if it is needed, I think using
```java
PropertyPlaceholderResolver getPropertyPlaceholderResolver() {
        if (this.propertyPlaceholderResolver == null) {
            this.propertyPlaceholderResolver = new DefaultPropertyPlaceholderResolver(new MapPropertyResolver(Collections.emptyMap()), ConversionService.SHARED);
        }
        return this.propertyPlaceholderResolver;
    }
```
will make `micronaut-openapi` compatible with all versions of `micronaut-core`.

alvarosanchez:
We're going to revert the change in core. See https://github.com/micronaut-projects/micronaut-core/issues/3181

However, we still want to do a new release of this module so that it is compiled with at least Micronaut 1.3.

graemerocher:
The change has been reverted

ilopmar (approved):

Issue #234, opened 2020-05-21
Kotlin suspend functions generate response schema of Object rather than specific type


csbiggar:
When using micronaut-openapi to generate a swagger spec for a kotlin suspend function, the response type is not being recognised, instead it's being generated as Object always. 

I think this is simlar to #209 where it's picking up on something kotlin-y under the covers?

Please see this fork for an example, details in the readme:
https://github.com/csbiggar/creating-your-first-micronaut-app-kotlin
(with thanks to [this pull request ](https://github.com/micronaut-projects/micronaut-openapi/pull/211) for the idea to borrow this project!)

Please could you take a look? Many thanks.

csbiggar:
Hello, thanks for investigating ... but ... I've upgraded to 1.5.1 and this is still not working. 

I've just pushed the upgrade to my example project above (and upgraded micronaut in it to 1.3.5 and kotlin to 1.3.72, just to be sure) , and you can follow the same steps to recreated in the readme and see it still failing. 

Could you take another look?

PR #235, opened 2020-05-23
Issue #234 - Handle return type for kotlin Continuation.

croudet:


graemerocher:
Thanks for looking at this @croudet 

Some of these APIs for coroutines I think we need in core, might be a good idea to send a PR to core to add APIs to check for coroutines because it is a general problem

croudet:
Currently the open-api still works with micronaut 1.x, so I have to implement workaround.

PR #242, opened 2020-06-02
Rapidoc focused render style

croudet:
Add new rapidoc FOCUSED render style

introduced in rapidoc 8.4.1
https://github.com/mrin9/RapiDoc/releases/tag/8.1.0


graemerocher:
Thanks for all the PRs!

PR #268, opened 2020-06-26
Changes Retention Policy of openapi annotation to SOURCE

croudet:
so that they are not kept in micronaut generated classes.

Requires micronaut-2.x


alvarosanchez:
Blocked by #223

croudet:
@ilopmar Thanks for taking a look at that.

ilopmar:
@croudet quick question before merging this. Is this a new feature, right? I'm asking to include it in 2.1.0 instead of 2.0.1

croudet:
It uses `TypedAnnotationTransformer` introduced in core 2.0.0, from the openapi point of view it could be considered as a new feature.

ilopmar:
Thanks, I'm going to update the `@since` tag to 2.1 and release a new version.

ilopmar (approved):

PR #277, opened 2020-07-05
Avoid to write multiple times the spec and view files.

croudet:
Happens with micronaut 2.0.
See https://github.com/micronaut-projects/micronaut-core/issues/3575


graemerocher:
Think we need to fix the upstream bug

PR #280, opened 2020-07-07
Groovy docs missing for views

harissecic:
Not sure if important but docs were missing info about Groovy+Gradle projects on how to generate views


graemerocher:
Thanks for the contribution!

PR #281, opened 2020-07-07
Groovy requires different scope

harissecic:
I played with Groovy project and OpenAPI and couldn't get swagger to generate until these 2 have been changed to compile

### DIFF
--- a/src/main/docs/guide/gettingStarted.adoc
+++ b/src/main/docs/guide/gettingStarted.adoc
@@ -2,13 +2,11 @@ To enable this support you should add the following dependencies to your build c
 
 dependency:micronaut-openapi[scope="annotationProcessor", version="{version}", groupId="io.micronaut.configuration"]
 for Groovy:
-dependency:micronaut-openapi[scope="compile", version="{version}", groupId="io.micronaut.configuration"]
+dependency:micronaut-openapi[scope="compileOnly", version="{version}", groupId="io.micronaut.configuration"]
 
 Then the following compile time dependency:
 
 dependency:swagger-annotations[scope="implementation", groupId="io.swagger.core.v3"]
-for Groovy:
-dependency:swagger-annotations[scope="compile", groupId="io.swagger.core.v3"]
 
 <1> The `openapi` configuration gets added into the annotation processor scope
 <2> The https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations[Swagger Annotations] are added to the compile classpath
--- a/src/main/docs/guide/gettingStarted.adoc
+++ b/src/main/docs/guide/gettingStarted.adoc
@@ -1,10 +1,14 @@
 To enable this support you should add the following dependencies to your build configuration:
 
 dependency:micronaut-openapi[scope="annotationProcessor", version="{version}", groupId="io.micronaut.configuration"]
+for Groovy:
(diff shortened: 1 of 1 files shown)


src/main/docs/guide/gettingStarted.adoc
```diff
@@ -1,10 +1,14 @@
 To enable this support you should add the following dependencies to your build configuration:
 
 dependency:micronaut-openapi[scope="annotationProcessor", version="{version}", groupId="io.micronaut.configuration"]
+for Groovy:
+dependency:micronaut-openapi[scope="compile", version="{version}", groupId="io.micronaut.configuration"]
```
ilopmar, line 5: This needs to be `compileOnly`

src/main/docs/guide/gettingStarted.adoc
```diff
@@ -1,10 +1,14 @@
...
 
 Then the following compile time dependency:
 
 dependency:swagger-annotations[scope="implementation", groupId="io.swagger.core.v3"]
+for Groovy:
+dependency:swagger-annotations[scope="compile", groupId="io.swagger.core.v3"]
```
ilopmar, line 11: Please use `implementation` instead of `compile`

src/main/docs/guide/gettingStarted.adoc
harissecic: Ok, may I ask why? Sice I just wanted it to start working I didn't go into details

src/main/docs/guide/gettingStarted.adoc
harissecic: Doesn't work in my case. Won't generate views

src/main/docs/guide/gettingStarted.adoc
ilopmar: Because for Groovy, the dependencies in the annotation scope processor only need to be `compileOnly`.

If you create a new application with `openapi` feature using https://launch.micronaut.io, you can check it:

image

After that, If you want the swagger-ui, you can add what you already submitted in this other PR: https://github.com/micronaut-projects/micronaut-openapi/pull/280/files

I've done that: Generate new project with Launch + adding the snippet in your PR, and it works as expected:

image

src/main/docs/guide/gettingStarted.adoc
ilopmar: Since Gradle 5 (I think) the recommendation is to use `implementation`, not `compile`. They are pretty much the same. Please google for differences about that if you are interested in learn more.

harissecic: > Since Gradle 5 (I think) the recommendation is to use `implementation`, not `compile`. They are pretty much the same. Please google for differences about that if you are interested in learn more.

I will retry this but as stated it does not generate views if using implementation.

ilopmar: Please see my other comment. It works.

harissecic: Apparently I left **annotationProcessor** for open API and that's the reason I didn't work, not the implementation on swagger. Sorry about that.

ilopmar:
Glad it worked for you. I'm closing the PR because we already have a note explaining the right scope for both Groovy and Kotlin.

image

Issue #279, opened 2020-07-07
Generated openapi yml file does not include all class fields after upgrade


runarhk:
### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Use generics for a dto used in a controller interface
2. Assemble project

### Expected Behaviour

All fields of the class should be included in schema.

### Actual Behaviour

The generic field is missing in produced openapi yml schema definitions

### Environment Information

- **Operating System**: Linux mint 19.3 Cinnamon
- **Micronaut Version:** 2.0.0, MN openapi 1.5.2
- **JDK Version:** openjdk 11.0.7 2020-04-14

### Example Application

https://github.com/runarhk/mn_openapi_generics_bug

runarhk:
This might be related. I am also having issues with getting the fields with type of an interface that is implemented by an enum.

Consider interface A has a field b of type B. B is an interface. C implements A, and field b of C is an enum D that imlements B.

alvarosanchez:
Fixed by #283

PR #283, opened 2020-07-07
Issue #279 - Refactored getBeanProperties

croudet:
Use different method to find fields and methods.


graemerocher:
Thanks for the contribution!

Issue #297, opened 2020-08-02
Parameters of type `Session`/`Session?` or those annotated with `@SessionValue` should be ignored


Enigmatrix:
# Steps to Reproduce
Create a controller with:
```kotlin
@Controller("/hello")
open class HelloController {
    @Get("/1")
    fun hello(session: Session): String {
        return "hello from session " + session.get("hello").get()
    }

    @Get("/2")
    fun hello2(@SessionValue hello: String): String {
        return "hello from session " + hello
    }
}
```
The generated swagger UI from the openapi specification is as follows:
![image](https://user-images.githubusercontent.com/12296254/89116119-0c66db00-d4c3-11ea-9fb5-6da914b53568.png)
![image](https://user-images.githubusercontent.com/12296254/89116128-26a0b900-d4c3-11ea-865a-be136898f732.png)

### Expected Behaviour

The parameters of type `Session`/`Session?` or those annotated with `@SessionValue` should be ignored as they are retrieved from the `SessionStore`, not from query parameters or body parameters. Additionally, these methods could be marked as Operations that require Cookie/Header (as per `HttpSessionConfiguration`).

### Actual Behaviour

They are present, and `Session` is considered an array of objects

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** 2.0.1
- **JDK Version:** 11

PR #298, opened 2020-08-04
Issue-297 - Parameters of type Session or those annotated with @SessionValue should be ignored

croudet:
https://github.com/micronaut-projects/micronaut-openapi/issues/297

Difficult to test because micronaut-openapi depends on an old version of micronaut-core


graemerocher:
Thanks for the contribution!

Issue #108, opened 2019-12-10
How to build Swagger/Rapidoc UI Views with build.gradle and Kotlin-only? 


lathspell:
The documenation on
https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html#openApiViews
has an example how to specifiy "-Dmicronaut.openapi.view.spec" in build.gradle file for a Java project.

This seems to have no effet on a Kotlin-only project though. I tried the following but with no success:

```
dependencies {
    ...
    kapt "io.micronaut.configuration:micronaut-openapi"
    implementation "io.micronaut:micronaut-views"
    implementation "io.swagger.core.v3:swagger-annotations"
    ...
}
compileKotlin {
    // https://kotlinlang.org/docs/reference/using-gradle.html
	kotlinOptions {
        verbose = true
	    jvmTarget = '1.8'
	    // Will retain parameter names for Java reflection
	    javaParameters = true
        // Generate OpenAPI UIs
        // DOES NOT COMPILE: freeCompilerArgs += '-Dmicronaut.openapi.views.spec=rapidoc.enabled=true,swagger-ui.enabled=true,swagger-ui.theme=flattop'
        freeCompilerArgs += '-Xjavac-arguments=-J-Dmicronaut.openapi.views.spec=rapidoc.enabled=true,swagger-ui.enabled=true,swagger-ui.theme=flattop'
        freeCompilerArgs += '-Xjavac-arguments=-Dmicronaut.openapi.views.spec=rapidoc.enabled=true,swagger-ui.enabled=true,swagger-ui.theme=flattop'
	}
}
```

How should I add those parameters? 

(BTW, compiler args seem somewhat ugly, can't this be configured just by adding a specific Gradle plugin or dependency?)

croudet:
Have you try with https://kotlinlang.org/docs/reference/kapt.html#java-compiler-options

lathspell:
I tried the following with no success. There is no error message either, the build runs fine but does not produce any HTML files or similar for the UI.

```
kapt {
    javacOptions {
        option('-J-Dmicronaut.openapi.views.spec', 'rapidoc.enabled=true,swagger-ui.enabled=true,swagger-ui.theme=flattop')
        option('-Dmicronaut.openapi.views.spec', 'rapidoc.enabled=true,swagger-ui.enabled=true,swagger-ui.theme=flattop')
[...]

PR #304, opened 2020-08-12
#108 Generating an OpenAPI view with Kotlin

sbeins55:
In a multi-module gradle project that used Kotlin with a kotlin script file instead of a groovy script file, I was able to successfully generate the Open API views when setting the `org.gradle.jvmargs` property. This allowed me to not have to use the environment variable approach to generate the views correctly. When trying to use the `kapt` block in my `build.gradle.kts`, I could not successfully generate the views.

### DIFF
--- a/src/main/docs/guide/openApiViews.adoc
+++ b/src/main/docs/guide/openApiViews.adoc
@@ -21,6 +21,12 @@ JAVA_TOOL_OPTIONS=-Dmicronaut.openapi.views.spec=redoc.enabled=true,rapidoc.enab
         ./gradlew --no-daemon clean assemble
 ----
 
+or in gradle.properties:
+[source,properties]
+----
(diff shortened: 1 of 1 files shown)


sbeins55:
This approach was taken from [a comment](https://github.com/micronaut-projects/micronaut-openapi/issues/108#issuecomment-584295590) on #108 by @croudet

graemerocher:
Is the correct approach to this not use KAPT annotation processor arguments? Fiddling with system wide gradle config seems an undesirable thing to document

sbeins55:
@graemerocher I wasn't able to generate the view resources with the KAPT annotation processor arguments. I was also able to get this working by using the `openapi.properties` file but that approach wasn't documented in this subsection. I can modify the PR to document the use of the properties file instead.

graemerocher:
Thanks for the contribution!

PR #324, opened 2020-09-09
Change the group id of the openapi dependency

sjrd218:
As of the 2.0.0 release it appears that the maven group was changed to io.micronaut.openapi.


ilopmar:
@sjrd218 Good catch. There are more places in the docs with the wrong group. Can you please update them too?

P.S: Also please sign the CLA.

sjrd218:
@ilopmar The other references looked like they were to the example snippets and looked appropriate. I used the Github edit feature to do this one, so there may indeed be other places that might need an edit.

ilopmar:
@sjrd218 Yeah, you're right. It would be good to move the examples to the same package, tho.

Issue #302, opened 2020-08-10
Use javax/jetbrains nullability annotations in required fields check for better Kotlin support


Panthaaaa:
When generating the required field list only javax.validation annotations are used:
https://github.com/micronaut-projects/micronaut-openapi/blob/c072c03355b417f36ff1a74c58bc557e0c4c52cf/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L748-L750

But when using Kotlin types are either nullable or not nullable which is represented by the javax.annotation.Nullable/Nonnull and org.jetbrains.annotations.Nullable/NotNull annotations. The Kotlin support could be improved massively by including these annotations in the generation. (Kotlin also has a reference for all known nullability annotations: https://github.com/JetBrains/kotlin/blob/master/core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JvmAnnotationNames.kt)

There are probably more parts of the generation relying on these annotations for required variables, they should also be enhanced.

graemerocher:
It shouldn't be necessary to add these since we map kotlin annotations to javax annotations. See https://github.com/micronaut-projects/micronaut-core/blob/master/inject/src/main/java/io/micronaut/inject/annotation/internal/KotlinNullableMapper.java

If this is not working, report an issue to https://github.com/micronaut-projects/micronaut-core/issues

graemerocher:
Oh I see the nullable ones are not handled. This seems like a simple fix, would you like to send a PR?

rohitkg98:
I would like to fix this. What do we mean by Nullable ones re not handled?

graemerocher:
this check: 

https://github.com/micronaut-projects/micronaut-openapi/blob/c072c03355b417f36ff1a74c58bc557e0c4c52cf/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L748-L750

Is not checking for `javax.annotation.Nonnull`

alvarosanchez:
Fixed via #325

PR #325, opened 2020-09-09
Checking Nonnull annotation for Kotlin compatibility

rohitkg98:
Fixed issue #302.
I was trying to add test. Should the functionality be that `nullable: false` to the generated schema?

If needed, I will make the required modifications.


graemerocher:
I believe it is `required: true`

rohitkg98:
> I believe it is `required: true`

As of the now, the functionality for all the annotations along with it is, that it sets `minLength: 1`. For Nonnull, should it set `required: true` or do the same?

graemerocher:
Thanks for the contribution!

Issue #286, opened 2020-07-08
Generate schema for inner classes with `dot` instead of `dollar`


ilopmar:
If I have the following POJO that I declare in my schema:

```java
public class SearchStatus {

    private String firstName;
    private String lastName;
    
    // constructor, getters, setters...

    public static class DocumentStatus {

        private String confirmationCode;
        private String filename;

        // constructor, getters, setters...
    }
}
```

The generated schema for the inner class uses `$` as separation:

![image](https://user-images.githubusercontent.com/559192/86898242-bc396a80-c108-11ea-92ce-1924c1be006e.png)

I think it would be nicer to use a `.`, like:

![image](https://user-images.githubusercontent.com/559192/86898345-e0954700-c108-11ea-8734-d3b1756e53a5.png)

PR #334, opened 2020-09-13
Generate schema for inner classes with `dot` instead of `dollar`

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/286

`$` is not allowed by OpenApi: `Component names can only contain the characters A-Z a-z 0-9 - . _`


graemerocher:
Thanks for all the PRs!

Issue #215, opened 2020-05-04
@ApiResponse overwrites default response


ftes:
The response generated by default, derived from the path methods return type, is helpful.

I want to add an  `@ApiResponse` annotation to describe an additional error response.

However, if I add that annotation, the generated openapi spec no longer contains the default response.

**Workaround:** Add an additional `@ApiResponse` annotation for the default response.

**Ideally:** The default response, derived from the method return type, is preserved (aka not overwritten) if I add an `@ApiResponse` annotation.

graemerocher:
I don't believe this is desirable behaviour. Being able to override is a feature and also useful on occasion when the autogenerated spec is not correct

aardee:
Problem with the @ApiResponse is that you cannot document Generic classes. My API returns ResponseWrapper<T>, which is processed correctly as default, but if I add ApiResponse for 400, I can no longer generate documentation for ResponseWrapper<T>.

mliberato:
Quoting `io.swagger.v3.oas.annotations.responses.ApiResponse.responseCode` property source code:

```java
public @interface ApiResponse {

    /**
     * The HTTP response code, or 'default', for the supplied response. May only have 1 default entry.
     *
     * @return response code
     **/
    String responseCode() default "default";

}
```


I think it should be a good ideia to allow this.
Most of the time, the autogenerated spec is OK using provided return type. It is true either you will always have an exceptional flow. So, for a standard method with a well defined return type, if you have a flow with just a different status code, you have to write all exceptions paths plus the original path.

The goal would be to write something as clear as this:

```kotlin
@ApiResponses(
        ApiResponse(),
        ApiResponse(
                description = "Pet not found",
                responseCode = "404"
        )
)
@Get("/pet/{id}")
fun detail(@PathVariable id: Long): Pet {
    TODO()
}
```

But you are forced to do this:

```kotlin
@ApiResponses(
        ApiResponse(
                responseCode = "200",
                content = [Content(
                        mediaType = "application/json",
                        schema = Schema(implementation = Pet::class)
                )]
        ),
        ApiResponse(
                description = "Pet not found",
                responseCode = "404"
        )
)
@Get("/pet/{id}")
fun detail(@PathVariable id: Long): Pet {
    TODO()
}
```

I think it should be straightforward to write only exceptional cases if you can. Note that this doesn't prevent one from overriding the default behaviour, as it is a matter of _not_ declaring an "empty/default" `ApiResponse`.

[...]

PR #342, opened 2020-09-17
Generate response mapping with status code instead of "default"

dstepanov:
https://swagger.io/docs/specification/describing-responses/ specifies that the status code should be the response key.

Also improves how `@ApiResponse` can be used, if one found with the same status code or `default` it will be used. 
The content is set if it's missing.

Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/215


dstepanov:
@graemerocher What is your opinion on this one?

adrianotr:
Thanks @dstepanov! Looking forward for this feature to be merged!
The current "default" response is making it difficult for us to document our REST endpoints that return 201 or other response statues.

graemerocher:
Thanks for the contribution!

JamesPace-GBI:
Hey, is it possible to remove the default 200 response from an API? We have an endpoint that returns a 201 i.e. never a 200, and we can't seem to remove it from the docs...

alvarosanchez:
@JamesPace-GBI thanks for the feedback. Contributions to the documentation are welcomed

Issue #343, opened 2020-09-17
Kapt is not working using "@Get(produces = APPLICATION_JSON, single = true)"


bj-eberhardt:
### Description

The gradle task kapt is not working if using annotations like `@Get(produces = APPLICATION_JSON, single = true)` and `@Post(consumes = ALL, produces = APPLICATION_JSON, single = true)`

### Steps to Reproduce

I created a new sample project using features "openapi" and "graphql" working with kotlin and gradle.
(see https://github.com/bj-eberhardt/micronaut-openapi-issue)

I had to override the GraphQLController and overwrite the "get" and "post" values.
That are annotated like this:

`@Get(produces = APPLICATION_JSON, single = true)` and `@Post(consumes = ALL, produces = APPLICATION_JSON, single = true)`

When running "gradle build --stacktrace" it fails at kapt with following stacktrace:

`Caused by: com.sun.tools.javac.processing.AnnotationProcessingError: java.lang.IllegalArgumentException: Invalid mime type: true
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(Unknown Source)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(Unknown Source)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(Unknown Source)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(Unknown Source)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(Unknown Source)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(Unknown Source)
        ... 34 more
Caused by: java.lang.IllegalArgumentException: Invalid mime type: true
        at io.micronaut.http.MediaType.<init>(MediaType.java:421)
        at io.micronaut.http.MediaType.<init>(MediaType.java:365)
        at io.micronaut.http.MediaType.of(MediaType.java:659)
        at io.micronaut.openapi.visitor.OpenApiControllerVisitor.lambda$null$0(OpenApiControllerVisitor.java:94)
        at io.micronaut.openapi.visitor.OpenApiControllerVisitor.lambda$mediaTypes$2(OpenApiControllerVisitor.java:94)
[...]

PR #344, opened 2020-09-19
Fix media type processing

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/343


graemerocher:
Thanks for the contribution!

PR #346, opened 2020-09-21
Allow defining @Schema at the field level with a custom implementation

dstepanov:


rdesgroppes:
Hi. Despite the good work, I happen to observe the following regression: properties definitions seem to now override components ones - for instance:
```java
@Schema(description="COMPONENT description")
public class Something {
}
...
@Schema
public class Else {
  @Schema(description="PROPERTY description")
  public Something something;
}
```
... now leads to sth like:
```
components:
  schemas:
    Something:
      description: PROPERTY description
```
... where it used to be:
```
components:
  schemas:
    Something:
      description: COMPONENT description
```
Do you guys observe that as well?

dstepanov:
Why would you define the schema at the class and the field level?
I don't think field schemas worked before anyway.

rdesgroppes:
Thanks for your reply, @dstepanov. [*Any sibling elements of a $ref are ignored*](https://swagger.io/docs/specification/using-ref/#sibling), right. What's new and surprising is the order of precedence when defined at both class and field levels, which practically leads _any_ field defined schema to supersede the class one. In that case (a redundant definition), I'd expected the class defined one to take precedence, as before, and any redundant field defined one to be ignored (maybe with a warning).

dstepanov:
Not sure exactly what are you trying to implement. But you are right the order is not the best.

Originally I added to allow overrinding the default schema generated for generic collections.

rdesgroppes:
> Not sure exactly what are you trying to implement.
Nothing special. Just realized that a project I was working on had redundant schema definitions (revealed by trying a local build from micronaut-openapi's master branch).

> Originally I added to allow overrinding the default schema generated for generic collections.
Might the recently merged fix for #367 also help? (by no longer generating component schemas s.a. `List_Pet_`)

dstepanov:
Propably yes, I think we also had problems with maps.

PR #359, opened 2020-10-06
Issue 3718 - changes to support registering classes output …

wetted:
Issue https://github.com/micronaut-projects/micronaut-core/issues/3718 - changes to support registering classes output generated resources so that interested downstream TypeElement visitors (like for graal) can post process them.

Fixing a bug where the JavaClassElementExt constructor didn't take into account the arrayDimensions property of the JavaClassElement argument

Update gradle wrapper to 6.7-rc-3

Update to Micronaut 2.1.0


gradle.properties
```diff
@@ -1,9 +1,9 @@
-projectVersion=2.1.1.BUILD-SNAPSHOT
+projectVersion=2.2.0.BUILD-SNAPSHOT
```
ilopmar, line 1: Please do not change this because if for some reason we need to revert the commit, we don't want to revert the project version. Once this is merged I'll update the project version in a separate commit (and also the core branch). 
Then a new 2.2.0 version of this module will be included in Micronaut 2.2.0

wetted: OK. I'll change it back and push the change.

ilopmar (approved):

Issue #156, opened 2020-02-22
Adding /login endpoint to Open API


zokica8:
Is there a way to add a /login endpoint generated by Micronaut itself with login enabled true in application.yml to Swagger Documentation?

andydoe19:
Hello @alvarosanchez If this is a duplicate, where can I find the answer please?

alvarosanchez:
@andydoe19 #197 duplicates #156

I don't know what "answer" are you looking for, but #156 is still open and therefore not implemented

JouperCoding:
+ 1 for this request in addition it would be nice to have /oauth/access_token & /logout added to the generated swagger.

dargenn:
Is there a way to manually add those entries?

ilopmar:
@dargenn Yes, you can use this: https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html#mergingSchemas

TorosyanV:
@alvarosanchez why you are marking it as duplicate, but still no answer?

alvarosanchez:
@TorosyanV you are probably misinterpreting the history of this issue.

This issue was never marked as duplicate. #197 was, as said in https://github.com/micronaut-projects/micronaut-openapi/issues/156#issuecomment-701951706

This issue was closed as fixed by #364

PR #364, opened 2020-10-15
Introduce @OpenAPIInclude to process already compiled classes for OpenApi

dstepanov:
Allows to include additional `@Controller` and `@Endpoint` classes that are already compiled.

Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/156


graemerocher:
Nice, maybe an example in the docs that shows how to include the login endpoints for security would be an idea?

Also an idea could be we include a meta annotation say `@OpenApiManagement` that is annotated with `@OpenAPIInlcude` and includes all the management endpoints to make it easier to include the default ones?

dstepanov:
Yeah, probably the same can be used for the security, I'm thinking it would be nice to support tags and security requirements on `@OpenAPIInclude`

graemerocher:
You could probably use an `AnnotationMapper` implementation to workaround it

dstepanov:
@graemerocher That works but any idea where to fix `AliasFor` on repeatable?

graemerocher:
Seems to be handled here https://github.com/micronaut-projects/micronaut-core/blob/2.1.x/inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java#L547

graemerocher:
Thanks for the contribution!

Issue #345, opened 2020-09-19
Add support for @RequestBean


asodja:
With Micronaut 2.0.0 there is new Bindable annotation `@RequestBean`. It looks like it does not work out of the box with OpenApi.

https://docs.micronaut.io/latest/guide/index.html#binding

`@RequestBean` maps bindable values to object. So object would probably need to be unwrapped in to OpenApi schema somehow.

morki:
The same fix is needed for "exploded operator" on get queries, example from documentation:

```java
@Get("/bookmarks/list{?paginationCommand*}")
public HttpStatus list(@Valid @Nullable PaginationCommand paginationCommand) {
    return HttpStatus.OK;
}
```

Now they are presented as object type.

PR #369, opened 2020-10-31
Add support for @RequestBean

asodja:
As title says this adds support for `@RequestBean` and closes https://github.com/micronaut-projects/micronaut-openapi/issues/345.

Fairly simple implementation: for every field of RequestBean class `processParameter` method is called. The only inconvenience is that I had to change some usages of `ParameterElement` to  `TypedElement`.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
@@ -42,8 +42,11 @@ import io.micronaut.http.uri.UriMatchTemplate;
 import io.micronaut.http.uri.UriMatchVariable;
 import io.micronaut.inject.ast.ClassElement;
 import io.micronaut.inject.ast.Element;
+import io.micronaut.inject.ast.FieldElement;
 import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.ast.ParameterElement;
+import io.micronaut.inject.ast.TypedElement;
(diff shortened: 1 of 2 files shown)


graemerocher:
Thanks for the contribution!

morki:
@asodja does this solve the expanded query parameters problem too?

https://github.com/micronaut-projects/micronaut-openapi/issues/345#issuecomment-716520475

asodja:
Hey @morki, mapping expanded query parameters to a pojo is a different functionality so this does not solve it for your example. I suggest you open new issue for that or maybe try to post a PR.

As a workaroud you can try with `@RequestBean`, this should work:
```
@Get("/bookmarks/list{?paginationCommand*}")
public HttpStatus list(@Valid @RequestBean PaginationCommand paginationCommand) {
    return HttpStatus.OK;
}
```

it works also without expanded parameters:
```
@Get("/bookmarks/list")
public HttpStatus list(@Valid @RequestBean PaginationCommand paginationCommand) {
    return HttpStatus.OK;
}
```

Note that `@RequestBean` might work a bit differently. Fields of `PaginationCommand` have to have `@QueryValue`  value annotation. Also since `@RequestBean` does not use any reflection, `PaginationCommand` has to be `@Introspected` and it must have a all fields constructor or setters for all fields or a static method with `@Creator`. `@Creator` also has to be applied to all args constructor if there are multiple constructors and all args constructor is not first.

morki:
@asodja It seems that `@RequestBean` is solving the same problem but elegantly and we can use it instead of expanded query params. Thank you very much for your contribution :)

asodja:
Awesome, that is great to hear. :)

PR #370, opened 2020-11-04
Allow Optional arguments to be recognised as non-required

bankras:
Currently Body arguments are accepting both @Nullable as Optional argument to be recognised as non required arguments.
For method @Parameters, this is only considered for @Nullable. With this fix, it will also accept Optional.


graemerocher:
Thanks for the contribution!

Issue #392, opened 2020-11-20
Improve support for @RequestBean, ignore static fields/constants. 


mikaelvik:
### Fix

This is (hopefully) an improvement to the solution from #369 . 

I've already found a simple solution and will submit a pull request. 

### Reproduce 

Create a @RequestBean class with a constant and a header field referencing the constant. 

```java
@Introspected
class MyRequestBean {

    public static final String HEADER_CONTENT_TYPE = "Content-type";

    @Header(HEADER_CONTENT_TYPE)
    private String contentType;

    // omitted
}
```

The generated openapi spec contains parameters for the constant and the field. 

```yml
# omitted ...
      parameters:
      - name: HEADER_CONTENT_TYPE
        in: query
        required: true
        schema:
          type: string
      - name: Content-type
        in: header
        required: true
        schema:
          type: string

```

### Expected Behaviour

The parameter `HEADER_CONTENT_TYPE` should not have been added.

mikaelvik:
I was able to work around the issue on my own project by moving the constants to a different class.

mikaelvik:
PR: https://github.com/micronaut-projects/micronaut-openapi/pull/394

It currently fails on something unrelated to the actual changes.

PR #393, opened 2020-11-20
Improved support for @RequestBean. Ignore static fields/constants (#392)

mikaelvik:
I changed the test code so that it defines the name of the header using a constant, the test failed. 

Ignoring static fields fixed the test.


mikaelvik:
Oops. Typo in the issue number.

PR #415, opened 2020-12-11
Add java.util.Optional support

pratik-brex:


openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -640,13 +640,18 @@ private Map toTupleSubMap(Object[] a, String entryKey, String entryValue) {
...
+            boolean isNullable = false;
 
             // StreamingFileUpload implements Publisher, but it should be not considered as a Publisher in the spec file
             if (!type.isAssignable("io.micronaut.http.multipart.StreamingFileUpload") && isContainerType(type)) {
                 isPublisher = type.isAssignable(Publisher.class.getName()) && !type.isAssignable("reactor.core.publisher.Mono");
                 isObservable = type.isAssignable("io.reactivex.Observable") && !type.isAssignable("reactor.core.publisher.Mono");
                 type = type.getFirstTypeArgument().orElse(null);
+            } else if (type.isAssignable("java.util.Optional") || type.isAssignable("com.google.common.base")) {
+               isNullable = true;
+               type = type.getFirstTypeArgument().orElse(null);
```
alvarosanchez, line 652: This seems it can be only once after the if/else block

alvarosanchez (changes requested):
Minor comment. Also, could you please add a test?

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
pratik-brex: sorry do you mean that we should recursively resolve it? so `Optional<Optional<X>> ` resolves to `nullable: true` and type `X` ?

alvarosanchez: No, I just mean the line `type = type.getFirstTypeArgument().orElse(null);`, which is duplicated at the end of both branches of the if/else, could be left after the if/else block.

pratik-brex: it's an `else if` so it wouldn't be correct to move it outside the block.

alvarosanchez: Ugh, you are right, ignore me

alvarosanchez (approved):

alvarosanchez:
Thanks for your contribution!

PR #433, opened 2021-01-01
#409 - Set consistent ordering for yaml map properties

### LINKED ISSUE #409: Generate endpoints and schemas in the same order into swagger.yml
We are using the generated OpenAPI spec in a custom manner, which means that the generated YAML is also included in our applications' repository to track the changes of it. This means that every time Micronaut regenerates the file, a custom Gradle task updates it in our repo as well. Because the generator doesn't order the endpoints and schemas by their name, two consequent run can end up in YAMLs that differ in the order of endpoints and schemas only.

### Task List

### Steps to Reproduce

1. Create a controller with multiple endpoints that have different paths (i.e. `/route-one` and `/route-two`)
2. Let Micronaut to generate the OpenAPI specs (swagger.yml) for the application again and again

### Expected Behaviour

The endpoints and schemas in the generated YAML should be in the same order for every run.

### Actual Behaviour

The endpoints and schemas are ordered randomly between different runs.

### Environment Information

- **Operating System**: MacOS Big Sur
- **Micronaut Version:** 2.2.1
- **JDK Version:** 11

I would be happy to contribute and fix this issue, but I'm quite unsure about where should I start it. I was browsing the source code, and it seems that sorting the endpoints (paths) would also solve the order of schemas implicitly.

asodja:
Fixes #409. 

Note: I first tried with: 
```
OpenApi sortOpenApi(OpenApi openApi) {
    yaml = serialize(openApi)
    return deserlialize(yaml)
}
```

But in that case some unit tests failed, since OpenApi ObjectMapper Deserializator does not handle some openApi cases that are written by hand in Unit tests. So some fields that Unit tests expect were not present in sorted deserialized OpenApi object. 

After that I decided to check actual raw yaml output in tests.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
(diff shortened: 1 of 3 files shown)


adamkobor:
Hey @asodja, thanks for your hard work, it looks good to me. However, I would extend the test case to see if the schemas are ordered consistently as well. What do you think?

graemerocher (approved):

asodja:
@adamkobor that sounds great. I invited you to my repo, so you can push branch to this branch. Or you can also create a separate push request after this one is merged.

adamkobor:
Hey @asodja, thanks, but I don't want to change anything about the logic, it seems OK for me. My intention was to add an other return type for example to the test case, and check if the schemas are ordered correctly as well (somewhere here: https://github.com/micronaut-projects/micronaut-openapi/pull/433/files#diff-440006bf8c9518f305809cbdfd2d9ff31be26b47e404a554368b6d52c485cd60R228)

alvarosanchez:
Thanks @asodja !

@adamkobor the test case you mentioned can be added in a separate PR if you still want to do it.

PR #446, opened 2021-01-25
Handle @Extension without name

### LINKED ISSUE #423: @Extension without name causes compilation error
### Task List

### Steps to Reproduce

See examples below.

### Expected Behaviour

When `name` is not provided for the `@Extenstion` annotation, it should use the default value of `""`

### Actual Behaviour

The compiler fails processing the open api configuration and fails to compile.

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** `2.2.0`
- **JDK Version:** `1.8`

### Compile Error
Omitting the `name` property produces compile error.

```java
        extensions = {
                @Extension(properties = @ExtensionProperty(name = "x-custom", value = "value"))
        }
```

### Workaround
Using the default allows it to succeed in compilation

```java
        extensions = {
                @Extension(name = "", properties = @ExtensionProperty(name = "x-custom", value = "value"))
        }
```

alvarosanchez:
Fixes #423


alvarosanchez:
@graemerocher the `@Extension` annotation has the `name` attribute defined as `String name() default ""`. Is it expected that `getRequiredValue()` fails in that case? Maybe it is, just to double check

graemerocher:
hmm we might normalise blank string to nothing, probably the correct fix either way

Issue #467, opened 2021-02-17
OpenAPI YAML file sorting POJO properties alphabetically


jacobmims:
This is likely a result of #433. While it is now consistent, this can result in awkward OpenAPI docs, where alphabetical sorting does not match the expected order of the properties.

### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create an object with properties not in alphabetical order
2. Generate the OpenAPI spec file
3. Properties in the spec file are in alphabetical order

### Expected Behaviour

Option 1: Unless specified, properties should be sorted in definition order. 

Option 2: Alphabetical is the default sort order, but provide support for `@Order` annotation.

### Actual Behaviour

Fields are ordered alphabetically regardless of definition order or usage of `@Order` annotation. 

### Environment Information

- **Operating System**: macOS 10.15.7
- **Micronaut Version:** 2.3.2
- **JDK Version:** 11.0.9.hs

### Example Application

[Example Repo](https://github.com/jacobmims/property-naming)

graemerocher:
hmm yeah not sure what we can do about this, the fix for #433 isn't easy to customise

asodja:
Yeah, maybe it's was not the best idea to modify ObjectMapper...

Original issue was, that paths and schemas had different order on different runs. So maybe it would be enough to just sort these two before putting out yaml. Smth like:
```
  // Sort paths
  Map<String, PathItem> sortedPaths = new TreeMap<>(openAPI.getPaths());
  openAPI.getPaths().clear();
  sortedPaths.forEach((k, v) -> openAPI.getPaths().put(k, v));
  
  // Sort schemas
  Map<String, Schema> schemas = openAPI.getComponents().getSchemas();
  openAPI.getComponents().setSchemas(new TreeMap<>(schemas));
  
  writeYaml(openApi);
```

jacobmims:
That sounds like a good compromise @asodja. I experienced the issue with the paths and schemas, so this fixed that issue beautifully. Impacting the properties seems like an unintended side effect, so if we can isolate the sorting to the paths/schemas, that would be the best of both.

asodja:
@jacobmims Could you maybe try my fix? How you can test it:
```
git clone git@github.com:asodja/micronaut-openapi.git
[...]

PR #468, opened 2021-02-17
Fix order of paths and schemas again

asodja:
This hopefully fixes ordering issue and also https://github.com/micronaut-projects/micronaut-openapi/issues/467.

It fixes order for Paths and all Components. Is there maybe any other case where some element that should be handled?


graemerocher:
Not that I am aware of. Thanks for the contribution

PR #473, opened 2021-02-22
Fix deprecation warnings

rdesgroppes:
This change proposes to address the few remaining deprecation warnings as suggested by micronaut API's javadoc:
* `ClassElement.getFields` -> `ClassElement.getEnclosedElements(ALL_FIELDS)`
* `JavaPropertyElement(..., ExecutableElement, ...)` -> `JavaPropertyElement(..., Element, ...)`
* `VisitorContext.visitMetaInfFile(String)` -> `VisitorContext.visitMetaInfFile(String, EMPTY_ELEMENT_ARRAY)`

The `-deprecation` option is only added to help locate future deprecation warnings.

### DIFF
--- a/openapi/build.gradle
+++ b/openapi/build.gradle
@@ -24,6 +24,8 @@ dependencies {
     }
 }
 
+compileJava.options.compilerArgs += '-deprecation'
+
 test {
     useJUnitPlatform()
 }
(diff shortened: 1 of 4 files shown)


graemerocher:
we need to consider that this will break using the module on earlier versions of Micronaut

PR #495, opened 2021-04-23
Upgrade dependencies: Micronaut, Groovy, Spock and target core 2.5.x

ilopmar:


ilopmar:
Already done as part of #497

Issue #483, opened 2021-03-28
Schema attributes with inheritance are dropped


swaechter:
### Steps to Reproduce

1.) Clone the example https://github.com/swaechter/openapiproblems
2.) Switch the version number in gradle.properties to 2.4.0 and run the application
3.) Check out the schema model `Owner` that has an attribute `vehicle` of the type Car/Bike
4.) Stop the application and update the version number 2.4.1. Run the application
5.) The schema attribute `vehicle` is now missing, but the schemas Car/Bike still exist. In addition, the Vehicle schema references Car/Bike via oneOf

### Expected Behaviour

The `vehicle` attribute shows up in the editor & is always referenced:

![MN3](https://user-images.githubusercontent.com/5626755/112749355-1f826a00-8fc2-11eb-8f77-a7d0305279ed.png)

![MN4](https://user-images.githubusercontent.com/5626755/112749358-21e4c400-8fc2-11eb-8ed7-c51d8dc818df.png)

### Actual Behaviour

Starting with 2.4.1, the `vehicle` attribute doesn't show up in the editor & isn't referenced anymore. I am not sure if this is somehow related to @JsonXXX attributes that aren't processed properly (Looking at https://github.com/micronaut-projects/micronaut-openapi/issues/471)

![MN1](https://user-images.githubusercontent.com/5626755/112749304-d6cab100-8fc1-11eb-8b7c-d4e1559dcb9d.PNG)

![MN2](https://user-images.githubusercontent.com/5626755/112749310-dcc09200-8fc1-11eb-855b-125212241d5d.png)

### Environment Information

- **Operating System**: Windows 10 x64
- **Micronaut Version**: Working in 2.4.0, broken in 2.4.1+
- **JDK Version**: AdoptJDK jdk-11.0.10.9-hotspot

### Example Application

https://github.com/swaechter/openapiproblems

swaechter:
It looks like the schema generation itself is correct. What's broken are the schema references (Schema X has an attribute Y of type Z --> Relationship Z has always the type/value "string" or "object" without $ref). I found these two broken relationships:

* anyOf: An Owner can jave a Car OR Bike, but it has to be of type Vehicle/contain all values of type Vehicle)
* $ref for nested classes or enums: ReadBackupSettingsDto has a backup frequency (Enum with weekday) and a time (Nested class with hour/minutes)

This means the generated OpenAPI code is unusable/broken as soon someone is using nested classes/enums :(

[...]

Issue #442, opened 2021-01-21
Not generating all fields from class when having children/sibling of same type


runarhk:
I have created a simple setup reproducing the where we have a class `a` that implements an interface `A` , having a field that has type `A` . 

This 

### Steps to Reproduce

1. Clone supplied repo
2. run gradle assemble
3. notice missing fields in type TestImpl1 in generated file at build/classes/java/main/META-INF/swagger/hello-world-0.0.yml

### Expected Behaviour

The field name `woopsie` should be in the TestImpl1 type with `type: $ref: '#/components/schemas/TestImpl1'`

### Actual Behaviour

The field `woopsie` is missing entirely. This is also the case for lists of same type. 

### Environment Information

- **Operating System**: Linux Mint
- **Micronaut Version:** TODO
- **JDK Version:** TODO

### Example Application

- TODO: link to github repository with example that reproduces the issue

graemerocher:
The field `woopsie ` is private, why would it be included in the public schema?

runarhk:
thanks for quick reply @graemerocher, i forgot to push the getter and setter which I added after removing lombok to get the test project as slim as possible. Pushed now ;)

alvarosanchez:
This happens because when the `whoopsie` field is visited, the `TestInterface` type is also being visited in progress, and there is a recursion break to stop processing it.

Not sure what solution could be applied here.

runarhk:
Makes sense. But how come it works if I add a generic type to the interface?

alvarosanchez:
I guess that in that case, it goes through a different code path since `TestInterface<T>` != `TestInterface`

PR #497, opened 2021-05-01
Fix issue where schemas with inheritance are dropped and also fix schema with recursion

asodja:
This is a proposal to:
Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/483
Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/442

Few notes:
- if property (getter) has custom type and type is referenced with $ref, then @Nullable or @Deprecated are not applied to `Schema` and you have to set them by hand via `@Schema`. I checked previous version and it seem that that did not work there before either. In that case schema should be copied and modified and saved under some different name.
- I added docs that `@Schema` on property is preferred over `@Schema` on type. And also that Schemas can be overriden if name is not defined.  That might be a bug, but I think it's still better to be documented, so users can find solution in docs.
- recursion fix also fixes correct generation of inheritance Schema. This can be seen in change in `OpenApiInheritedPojoControllerSpec.groovy`. For example for class `class Cat extends Pet {...}`, previous yaml was:
```
    Cat:
      type: object
      properties:
        clawSize:
          type: integer
          description: ""
          format: int32
```
But not is:
```
    Cat:
      allOf:
      - $ref: '#/components/schemas/Pet'
      - type: object
        properties:
          clawSize:
            type: integer
            description: ""
            format: int32
```


ilopmar:
Thanks!

Issue #361, opened 2020-10-12
Duplicate operationId values


davidhiendl:
Currently the operationId for the OpenAPI spec is generated only from the simple method name (if not explicitly set).
This basically forces the use of `@Operation(operationId = "....")` for every non-unique endpoint method name to avoid conflicts in the generated spec.
This results in the need to add many additional annotations to resolve conflicts.

### Steps to Reproduce
Create 2 classes containing a identically named method without a `@Operation(operationId = "....")` annotation

```kotlin
@Controller("/test1")
class TestController {
    @Get("/")
    fun index(): String {
        return "Hello World"
    }
}

@Controller("/test2")
class TestController {
    @Get("/")
    fun index(): String {
        return "Hello World"
    }
}
```

After building the project both operations have the same operationId in the resulting OpenAPI spec.

### Expected Behaviour
It would be preferable if a unique operationId is generated from the full method path and HTTP method.
```yml
paths:
  /test1:
    get:
      operationId: test1Get
      ....
  /test2:
    get:
      operationId: test2Get
      ....
```

### Actual Behaviour
operationId is duplicate because only the method name is used when generating the operationId:

```yml
paths:
  /test1:
    get:
      operationId: index
      ....
  /test2:
    get:
      operationId: index
      ....
```

The relevant code is located here:
https://github.com/micronaut-projects/micronaut-openapi/blob/1f6c6e8560a6976e21a2c7a569885db7090408f1/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java#L728

### Environment Information
- **Micronaut Version:** 2.1

Tomucha:
... or maybe possibility to inject "operationid naming strategy".

Tomucha:
Just to clarify a bit further - it often isn't possible to simply rename the method or add an annotation. Controllers might have common ancestor, in which case I have no possibility to influence resulting 'operationId' of the shared ancestor method and I end up with invalid OpenAPI specification ... (or is there something I can do?)

davidhiendl:
@Tomucha I worked around your problem with common methods by implementing them without annotations in the super class and then simply overwriting them and adding annotations as needed and calling the parent method.

Tomucha:
That's sounds like a lot of "super" calls, but I don't see any other possibility either. Thanks!

ilopmar:
[...]

PR #498, opened 2021-05-01
Generate unique operation ids

asodja:
Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/361

Logic:
- Operation id is generated as: `javaMethodName + HttpMethodName`, e.g. `@Get void index()` will produce `indexGet`.
- if there is any duplicate found a number is added, for example if user have two `index` methods, one will be `indexGet1` and other will be `indexGet2`
- for user defined operation ids this logic is not applied, only warning is logged if there is any duplicate found
- also conflicts between user operation ids and generated ids are resolved


davidhiendl:
Does this really resolve issue #361 because from what I can see in the commit (correct me if I'm wrong) the operation ID does still not include anything to "qualify" operation IDs with their controllers?

asodja:
Hey David, 

I believe the main issue was that operation ids are duplicated: that is what this PR solves. There was very little written about qualifying operation IDs with their controllers or maybe I missed it.

Feel free to improve/change solution with a PR or file/reopen a issue.

Note: I am not a Micronaut developer.

swaechter:
Hi @asodja  I just updated an application up to Micronaut 3.0.0 M5 and run into the issues of changed operation IDs. I think your PR is a breaking change and needs some mentioning/comment in a release note - otherwhise people think this is a bug (I am soo used that method name = operation ID, I never specify the operation ID. But I never thought about method overloading in Java that might break the generation. Good catch!)

swaechter:
Some additional thoughs:

* This change affects developers/companies that only provide a backend, but no consumer (Frontend or test client) that generates and consumes the OpenAPI definition from the backend/OpenAPI file
  * Many (if not all) developers/companies only do endpoint testing and don't use the generated client. Looking at #517 (Would be awesome)
  * Thus the problem will be found by accident or complaining API consumers
* Developer/companies (like me) that have a frontend (React + Typescript + OpenApiGenerate plugin to generate a `typescript-axios` that is built together will run into a broken build (E.g getAccounts changed to getAccountsGet. Compared to Quarkus/Spring that's quite unexpected and typing out the method name again feels strange --> Makes refactoring harder)
* In addition all exposed OpenAPI examples in https://micronaut-projects.github.io/micronaut-openapi/latest/guide/index.html need to be updated

ilopmar:
@swaechter Can you please explain what you should expect would be the fix for the original issue?

If you have

```
@Controller("/test1")
class TestController {
    @Get("/")
    fun index(): String {
        return "Hello World"
    }
}

@Controller("/test2")
class TestController {
    @Get("/")
    fun index(): String {
        return "Hello World"
    }
}
```

What would be the expected openapi.yml generated?

swaechter:
Hi Iván. To be honest I never thought about methods with the same name. I just never encoutered that situation.

I don't know what a good strategy is, My personal preference would be to throw a compile time error to keep the implementation/strategy simple + ask the user to rename the method or use `@Operation(operationId="....")`.

I was just concerned about developers that upgrade and are not aware of that change - it can/will break their OpenAPI schema.

ilopmar:
Thanks for the quick reply. I'm not sure either what should be the best solution to this. Throwing an exception maybe it's a little bit too much because maybe there are other options (the controller comes from other library,...)

I'll try to see what Spring does in cases like this so we can learn what they decided. In any case I think I'm going to revert that change, because as you said is a breaking change and for not a good reason.

Thanks for your input.

ilopmar:
I've investigated what Spring Boot does and I have to say it makes way more sense. When there is a duplicate operationId they add a suffix: `_1`, `_2`,... With this approach, by default and in 99% of the cases nothing will change because the `operationId` generated is the name of the method. Only in those cases where there is an overlap the suffix is added.

I'll work on making these changes.

PR #499, opened 2021-05-03
fixes #363 adds support for schema READ_ONLY and WRITE_ONLY accessModes

### LINKED ISSUE #363: Add support for @Schema(accessMode = ...) in classes
It would be useful to take into account the [`accessMode`](https://docs.swagger.io/swagger-core/v2.1.5/apidocs/io/swagger/v3/oas/annotations/media/Schema.html#accessMode--) element of `@Schema` in classes.

This way some fields could be specified only for requests or responses. For instance:
```java
@Schema(name="MyPet", description="Pet description")
class Pet {
    private int age;
    private String name;

    public void setAge(int a) {
        age = a;
    }

    @Schema(description="Pet age", maximum="20", accessMode = Schema.AccessMode.WRITE_ONLY)
    public int getAge() {
        return age;
    }

    public void setName(String n) {
        name = n;
    }

    @Schema(description="Pet name", maxLength=20, accessMode = Schema.AccessMode.READ_ONLY)
    public String getName() {
        return name;
    }
}
```
Given the definition above: `age` would only appear wherever `Pet` is part of a request body, whereas `name` would only appear wherever `Pet` is part of a response body.

LajosPolya:
This PR fixes issue #363. I added support for READ_ONLY and WRITE_ONLY accessModes in the Schema annotation.

Setting accessMode to READ_ONLY sets the schema's readOnly attribute to true.
Setting accessMode to WRITE_ONLY sets the schema's writeOnly attribute to true.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -23,7 +23,6 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
-import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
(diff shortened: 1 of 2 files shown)


graemerocher:
Should we be automatically populating this is the property itself has no setter?

el-duderino5:
Is there any ETA on this PR?

graemerocher:
Thanks for the contribution!

LajosPolya:
I see it's merged now but looking at the code I couldn't see a better way of doing it.
Sorry for the late response

PR #511, opened 2021-05-23
Remove swagger-annotations version

pehecoro:
Dependency swagger-annotations version of io.swagger.core.v3 isn't the same version as Micronaut. 
If you select version 2.5.0 in the docs, for example, there's no 2.5.0 of swagger-annotations in the maven repository.


ilopmar:
The right fix for this is to remove the version because we already include it in Micronaut's BOM:

- 2.5.x: https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/gradle.properties#L96
- 3.0.x: https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/gradle.properties#L95

Can you please amend this PR and apply the changes in #512 here too?

pehecoro:
Oh, thanks!

Sure, no problem! Removed both #511 and #512 here.

ilopmar:
Thanks!

PR #530, opened 2021-07-19
Upgrade to Micronaut 3.0

ilopmar:
When CI passes I'll use Micronaut 2.5.9 version but still target Micronaut 3.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,8 +1,8 @@
-projectVersion=2.5.1-SNAPSHOT
+projectVersion=3.0.0-SNAPSHOT
 jacksonVersion=2.12.2
 micronautDocsVersion=2.0.0.RC1
-micronautVersion=2.5.3
-micronautTestVersion=2.3.3
+micronautVersion=3.0.0-M2
+micronautTestVersion=2.3.7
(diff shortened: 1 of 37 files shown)


graemerocher:
I tested the existing release with Micronaut 3.0 snapshots and it seems to work fine

ilopmar:
The problem of the existing release is the findbugs dependency. I did this https://github.com/micronaut-graal-tests/micronaut-basic-app/blob/3.0.x/build.gradle#L41-L42 in the GraalVM test app as a workaoround.
In this PR I remove that class so it works without adding additional dependencies to Micronaut 3.0 apps.

ilopmar:
Ummm.... now that I think about it probably we don't need a new major version, right @graemerocher? Probably a new minor one would be enough.

graemerocher:
yup

sdelamo (approved):
Please, include this in the commit message: 

- micronaut from 2.5.3 to 2.5.9
- micronaut test from 2.3.3 to 2.3.7
- swagger from 2.1.9 to 2.1.10
- fix nullability annotations
- replace javax.inject with jakarta.inject

PR #539, opened 2021-08-02
Generate correct openapi when returning a StreamedFile

### LINKED ISSUE #441: Returning StreamedFile does not fulfill OpenAPI specs
A controller function returning a `StreamedFile` does not generate a OpenAPI.yml according to the OpenAPI 3.0 specs. It instead describes the fields etc of the `StreamedFile` class.

### Steps to Reproduce

```
    @Produces("application/pdf")
    @Get("/{documentId}")
    fun getOne(documentId: UUID): HttpResponse<StreamedFile> {
        [...]
```
### Expected Behaviour
The [OpenAPI Specs](https://swagger.io/docs/specification/describing-responses/#response-that-returns-a-file) specifies files like this:
```
  /v1/documents/{documentId}:
    get:
      [...]
      default:
        description: getOne default response
        content:
          application/pdf:
            schema:
              type: string
              format: binary
```

### Actual Behaviour
The code above produces this `OpenAPI.yml`
```
  /v1/documents/{documentId}:
    get:
      [...]
      responses:
        default:
          description: getOne default response
          content:
            application/pdf:
              schema:
                $ref: '#/components/schemas/StreamedFile'
```

### Environment Information

- **Micronaut Version:** 2.1
- **JDK Version:** 1.8

ilopmar:
Fixes #441

### DIFF
--- a/openapi/build.gradle
+++ b/openapi/build.gradle
@@ -6,6 +6,8 @@ dependencies {
     compileOnly "io.micronaut:micronaut-inject-java"
 
     implementation "io.micronaut:micronaut-http"
+    implementation "io.micronaut:micronaut-http-server"
+
     api "io.swagger.core.v3:swagger-core:$swaggerVersion"
     api "io.swagger.core.v3:swagger-models:$swaggerVersion"
     api "io.swagger.core.v3:swagger-annotations:$swaggerVersion"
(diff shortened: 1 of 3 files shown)


ilopmar:
Please wait, I still need to do add a test when the controller is annotated with `@Schema`

ilopmar:
@jameskleeh ready for review

PR #540, opened 2021-08-04
All not nullable parameters in a constructor are required in OpenAPI

### LINKED ISSUE #409: Generate endpoints and schemas in the same order into swagger.yml

ilopmar:
Fixes #409

It turns out that at the end this is not a breaking change. As you can see in the modified tests I only needed to make some minor adjustments, so I think this fix can go in a patch release.

---

According to https://swagger.io/docs/specification/data-models/data-types/#required
all object properties are optional by default, so if a fields is not
nullable, it should be declared as "required" in the OpenAPI spec file.

In case the `@Schema` annotation exists on the field, it takes
precendece because the user decided to use it. It's up to the user to
make sure that the annotations in the field and in the constructor are
the same so they make sense.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -25,10 +25,10 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.micronaut.annotation.processing.visitor.JavaClassElement;
 import io.micronaut.annotation.processing.visitor.JavaClassElementExt;
 import io.micronaut.core.annotation.AnnotationClassValue;
 import io.micronaut.core.annotation.AnnotationValue;
-import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.beans.BeanMap;
 import io.micronaut.core.bind.annotation.Bindable;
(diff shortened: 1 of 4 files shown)


ilopmar:
After discussing this with @jameskleeh we've agreed that this is a breaking change. I'll merge them after the other PRs are merged and it will be included in a next major version targeting Micronaut 3.0.x

jgoldman-nydig:
@ilopmar after upgrading to micronaut `2.5.12` I'm seeing properties that used to be marked as required no longer being marked as such, for example the following kotlin data class:

```
    @Schema(name = "AuthorizationResponse")
    data class AuthorizationResponse(
        val username: String
    )
```

This doesn't sound like the intended functionality based on this PR.

jonnii:
@ilopmar after upgrading to micronaut `2.5.12` I'm seeing properties that used to be marked as required no longer being marked as such, for example the following kotlin data class:

```
    @Schema(name = "AuthorizationResponse")
    data class AuthorizationResponse(
        val username: String
    )
```

This doesn't sound like the intended functionality based on this PR.

jameskleeh:
@jonnii The changes in this PR should not be in the version of micronaut-openapi included in 2.5.12. If you are using version 3.0.0 then I would expect `username` to be required.

jonnii:
@jameskleeh ok so then in that case upgrading to 2.5.12 is a breaking change for us =( would it be possible to backport this to 2.5.x?

jameskleeh:
@jonnii Please create a new issue with an example project. If there were breaking changes introduced then they should be reverted. This PR cannot be back-ported because it contains breaking changes

jonnii:
@jameskleeh happy to be obliged! 

https://github.com/jonnii/demo/commit/d30b76ba0a73ef669822cb9f507e990efd1c44b4

This diff shows an upgrade from `.8` to `.12` and the corresponding swagger change.

Issue #577, opened 2021-09-24
Interface Inheritance


cemo:
### Expected Behavior

OpenApi spec must be created properly for interfaces too. 

### Actual Behaviour

It is working for classes but not interface based inheritance.

### Steps To Reproduce

1. checkout main branch
2. clone https://github.com/cemo/swagger-spec
3. mvn clean compile

Check `target/classes/META-INF/swagger/swagger-spec-0.0.yml`

It has **not including** `BaseQueryParams`

-----

switch into `working-properly`

Check `target/classes/META-INF/swagger/swagger-spec-0.0.yml`

It has **including** `BaseQueryParams`

### Example Application

https://github.com/cemo/swagger-spec

### Version

3.0.3

graemerocher:
`@JsonProperty` is not an inherited annotation

cemo:
@graemerocher  How can I make it inherited?

cemo:
I cleaned all jsonProperty annotations. Result is same. Actually It was simplified version piece of code but I forgot to delete that annotations.

cemo:
`  - $ref: '#/components/schemas/BaseQueryParams'` is missing in the first version. I can take a look if you guide me @graemerocher.

```
openapi: 3.0.1
info:
  title: swagger-spec
  version: "0.0"
paths:
  api/products:
    get:
      operationId: getProducts
      parameters: []
      responses:
        "200":
          description: getProducts 200 response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ProdcatProductsResponse'
components:
  schemas:
    BaseQueryParams:
      type: object
      properties:
        locationId:
          type: string
        productId:
          type: integer
          format: int32
    ProdcatProductsResponse:
      type: object
      properties:
        searchMetadata:
          $ref: '#/components/schemas/ProdcatSearchMetadata'
    ProdcatSearchMetadata:
      allOf:
      - $ref: '#/components/schemas/BaseQueryParams'
      - type: object
        properties:
          page:
            $ref: '#/components/schemas/ProdcatSearchMetadata.Page'
    ProdcatSearchMetadata.Page:
      type: object
      properties:
        totalPages:
          type: integer
          format: int32
        totalResults:
          type: integer
          format: int32

```

cemo:
@graemerocher 
[...]

PR #581, opened 2021-09-28
Upgrade to Micronaut 3

cemo:
This PR is sent separated to address #577.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
@@ -773,7 +773,10 @@ abstract class AbstractOpenApiEndpointVisitor extends AbstractOpenApiVisitor {
     private ClassElement returnType(MethodElement element, VisitorContext context) {
         ClassElement returnType = isJavaElement(element.getOwningType(), context) ? JavaClassElementExt.getGenericReturnType(element, context)
                 : element.getGenericReturnType();
-        if (returnType.isAssignable(void.class) || returnType.isAssignable("io.reactivex.Completable")) {
+        if (returnType.isAssignable(void.class) || returnType.isAssignable("io.reactivex.Completable")
+            || (returnType.isAssignable("reactor.core.publisher.Mono")
+                && returnType.getFirstTypeArgument().isPresent()
+                && isVoid(returnType.getFirstTypeArgument().get()))) {
             returnType = null;
         } else if (isResponseType(returnType)) {
             returnType = returnType.getFirstTypeArgument().orElse(returnType);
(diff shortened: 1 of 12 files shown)


cemo:
The PR is failing on my local environment. I am trying to understand what is wrong but the response 

` $ref: #/components/schemas/MutableHttpResponse_Object_` 

It seems it is not working properly on 3.x. I am looking for it now.

ilopmar:
@cemo You need to convert the project to be Micronaut 3 compatible, check this comment: https://github.com/micronaut-projects/micronaut-core/issues/5607#issue-922763986

cemo:
It was actually a just a simple PR to address issue we hit. Now I am in the process of migrating to 3.x. 😂

cemo:
@ilopmar I checked that PR. This project somehow has already done these changes but it is only missing these props I changed. Changing these props is causing a test case to fail. Would you please provide necessary guidance about this project where should I take a look?

ilopmar:
The tests are failing because of missing RxJava2 classes, you need to add the dependency:

```groovy
testImplementation "io.reactivex.rxjava2:rxjava"
```

Also it is probably a good idea to add some test that uses `Mono` instead of `Single`/`Maybe`. For that you also need to add Project Reactor core dependency.

cemo:
@ilopmar 

When I debug, I see that the implementation has changed slightly.

https://github.com/micronaut-projects/micronaut-openapi/blob/aab3652dd723d48b569ad9a8e3f40e673099b489/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L1360-L1364

in 2.x returns "Object"
in 3.x returns "MutableHttpResponse_Object_"

cemo:
https://github.com/micronaut-projects/micronaut-openapi/blob/aab3652dd723d48b569ad9a8e3f40e673099b489/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L1346-L1358

How should be the correct implementation of this with 3.x? 
For `io.micronaut` prefixed packages returning `Object` sounds ridiculous? 

cc @graemerocher

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
```diff
@@ -888,9 +891,15 @@ private boolean isResponseType(ClassElement returnType) {
     }
 
     private boolean isSingleResponseType(ClassElement returnType) {
-        return returnType.isAssignable("io.reactivex.Single")
-                && returnType.getFirstTypeArgument().isPresent()
-                && isResponseType(returnType.getFirstTypeArgument().get());
+        boolean assignable = returnType.isAssignable("io.reactivex.Single") ||
+                             returnType.isAssignable("org.reactivestreams.Publisher");
```
cemo, line 895: This part was the actual problem

cemo:
@ilopmar I addressed remaining issues. I think we are ready to go. Would you please review it? I will send another PR to address another issue as well after it.

ilopmar (approved):

ilopmar:
Thanks for your contribution!

PR #588, opened 2021-10-03
Set required correctly when required property is not explicitly set

### LINKED ISSUE #587: Non required fields are incorrectly marked as required in the swagger ui
### Expected Behavior

the default value for required is false in the @Schema annotation, hence the generated swagger ui documentation should show respective fields as not required (without red asterisk) 
@Schema(description = "some description", implementation = Foo.class)

When the required value is explicitly set to false then the generated swagger ui documentation should show respective fields as not required (without red asterisk) 
@Schema(description = "some description", implementation = Foo.class, required = false)

### Actual Behaviour

The fields with default value for required in @Schema annotation are shown as required (with red asterisk) in the swagger UI

The fields with explicit false value for required in @Schema annotation are shown as not required (without red asterisk) in the swagger UI, which is expected.

### Steps To Reproduce

Add an annotation on a field
@Schema(description = "some description", implementation = Foo.class, required = false)
and run the application.
Access the swagger ui documentation.

### Environment Information

MAC
JDK16

### Example Application

_No response_

### Version

3.0.1

sujeet100:
Set required correctly when required property is not explicitly set in the @Schema annotation
Fixes #587

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -807,7 +807,7 @@ abstract class AbstractOpenApiVisitor  {
     private Optional<Boolean> hasElementSchemaRequired(Element element) {
         AnnotationValue<io.swagger.v3.oas.annotations.media.Schema> schemaAnnotationValue = element.getAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
         if (schemaAnnotationValue != null) {
-            return schemaAnnotationValue.booleanValue("required");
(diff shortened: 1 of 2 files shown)


ilopmar:
The way to define a field as optional is annotating it as `@Nullable` in the constructor. It doesn't really make sense to define it as non nullable (not defining the `@Nullable` annotation) but then mark it as `required = false` in the OpenAPI.

In any case I think your fix makes sense because the configuration defined in the `@Schema` annotation should take precedence when defined. It's "up to the user" to make sure that the nullability makes sense in the OpenAPI.

Please sign the CLA and write a proper test. The one you added passes without applying your code changes. You can use the same in the linked project you shared in this issue.

sujeet100:
Thanks for the feedback. I have fixed the test case. Also signed the CLA.

Issue #592, opened 2021-10-08
Support Kotlin return type Unit 


morki:
### Feature description

Just like #591 but for Kotlin's equivalent `kotlin.Unit`.

morki:
I think it can be as easy as adding `kotlin.Unit` here:

```java
private boolean isVoid(ClassElement returnType) {
    return returnType.isAssignable(void.class) || returnType.isAssignable("java.lang.Void");
}
```

I'll try to create PR.

PR #597, opened 2021-10-15
Support kotlin's Unit as return type

ashwini-desai:
Fixes [this](https://github.com/micronaut-projects/micronaut-openapi/issues/592) issue.


graemerocher:
We only recently worked out how to compile kotlin code dynamically in tests. This is available using `AbstractKotlinCompilerSpec` https://github.com/micronaut-projects/micronaut-core/blob/3.1.x/inject-kotlin-test/src/test/groovy/io/micronaut/annotation/processing/test/Test.groovy

It is not used in this project yet though

ilopmar:
@ashwini-desai done!

PR #606, opened 2021-11-01
Change nullable handling

### LINKED ISSUE #605: Nullable properties generate incorrect OpenAPI specification
### Expected Behavior

I updated a project from micronaut 2.5.11 to 3.1.2 and the typescript SDK generator we use stoped working correctly.

The problem is related to how nullable properties are handled. 

Given a class with a nullable property like this
```
data class ParentSchema(
  val child: ChildSchema? = null
)
```

That class would procude this in micronaut 2.5.11 :
```
  ParentSchema:
      type: object
      properties:
        child:
          $ref: '#/components/schemas/ChildSchema'
```

As of micronaut 3.1.2, it will produce :
```
  ParentSchema:
      type: object
      properties:
        child:
          allOf:
          - $ref: '#/components/schemas/ChildSchema'
          - type: object
            nullable: true
```
This particular structure cause generators to see this as being a ChildSchema or an object, not a nullable property.

Instead, it should produce something like this :
```
In OAS 3.0
  ParentSchema:
      type: object
      properties:
        child:
          nullable: true
         oneOf:
         - $ref: '#/components/schemas/ChildSchema'

In OAS 3.1
  ParentSchema:
      type: object
      properties:
        child:
          oneOf:
            - 'null'
            - $ref: '#/components/schemas/ChildSchema'
```

You can find much more details about nullable properties [here](https://github.com/OpenAPITools/openapi-generator/issues/5180). 

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.1.2

Chris-V:
First commit changes the output for `@Nullable` properties as per OAS 3.0:
```
ParentSchema:
  type: object
  properties:
    child:
      nullable: true
      oneOf:
        - $ref: '#/components/schemas/ChildSchema'
```

Second commit sets `nullable: true` for `Optional<>` fields. I felt like this was desirable, but I can strip that commit if not.

I believe both commits are breaking changes.

Fixes #605

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -978,7 +978,7 @@ abstract class AbstractOpenApiVisitor  {
         }
         // @Schema annotation takes priority over nullability annotations
         Boolean isSchemaNullable = element.booleanValue(io.swagger.v3.oas.annotations.media.Schema.class, "nullable").orElse(null);
-        boolean isNullable = (isSchemaNullable == null && element.isNullable())
+        boolean isNullable = (isSchemaNullable == null && (element.isNullable() || isTypeNullable(elementType)))
             || Boolean.TRUE.equals(isSchemaNullable);
         if (isNullable) {
             topLevelSchema.setNullable(true);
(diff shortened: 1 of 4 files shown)


ilopmar:
Thanks for the PR. I'll take a look at it soon.

Chris-V:
Any chance to get this reviewed please? :)

Chris-V:
@ilopmar Is this desirable? I will stop maintaining the PR if this is not something you are interested with

denebgarza:
I was about to open an issue about `java.util.Optional` properties not being set as `nullable: true`. I think this is desirable, and what I would expect should happen based on the [isTypeNullable](https://github.com/micronaut-projects/micronaut-openapi/blob/891b222c1b370dc1848864d34aca18980d0eac0e/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L609-L611) method.

graemerocher:
Thanks for the contribution!

PR #607, opened 2021-11-03
Do not use internal classes from micronaut-core

### LINKED ISSUE #593: Do not use Micronaut core `@Internal` classes
Currently in `JavaClassElementExt` and `JavaMethodElementExt` we are extending from micronaut core `@Internal` classes `JavaClassElement` and `JavaMethodElement`.

This makes this module fragile because when we modify `JavaClassElement` constructors this module doesn't compile anymore.

Analyze why those classes are needed and remove them. Maybe some changes are needed in core to expose some missing information about generics.

ilopmar:
Fixes #593

This PR removes the usage of Core's `@Internal` class `JavaClassElement`. We still have our own `JavaClassElementExt` but it doesn't extends from anything. That class has a lot of code copied from `JavaClassElement`because it is necessary to initialize the bean properties defined in a "fluent way" (like when using Lombok). See https://github.com/micronaut-projects/micronaut-core/pull/6415 for more information.
As suggested in that PR we could [explore another approach](https://github.com/micronaut-projects/micronaut-core/pull/6415#issuecomment-951079610) instead of adding another public method to `ClassElement`. We will explore that in the future, but at least, even with the code duplication here, we don't depend on the internal class anymore.


openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiIncludeVisitor.java
```diff
@@ -67,9 +66,8 @@ public void visitClass(ClassElement element, VisitorContext visitorContext) {
         }
     }
 
-    private void visit(TypeElementVisitor visitor, VisitorContext visitorContext, ClassElement ce) {
+    private void visit(TypeElementVisitor<?, ?> visitor, VisitorContext visitorContext, ClassElement ce) {
         visitor.visitClass(ce, visitorContext);
-        JavaClassElementExt javaClassElement = new JavaClassElementExt(ce, visitorContext);
-        javaClassElement.getCandidateMethods().forEach(method -> visitor.visitMethod(method, visitorContext));
+        ce.getEnclosedElements(ElementQuery.ALL_METHODS).forEach(method -> visitor.visitMethod(method, visitorContext));
```
graemerocher, line 71: This not filtering out static & private methods like the previous code

graemerocher (changes requested):
probably and improvement but it still depends on `JavaMethodElement` which is internal API

ilopmar:
I've added the missing filters (although I've duplicated the filter code in both places)

I missed the `JavaMethodElement`. It's not urgent because every time this has been broken has been because of `JavaClassElement`. In any case, I'll explore the approach suggested in the PR I sent to core. If we add the missing methods/filters/... to core to get the appropriate methods here, then we could get rid of completely of `JavaClassElementExt` in this module and use no internal classes.

graemerocher:
This is not quite what I meant, you need to use the `ElementQuery` API for the filtering, we can talk about it tomorrow

graemerocher (approved):

PR #621, opened 2021-12-10
Upgrade to internal build plugins 5.0.3

ilopmar:


docs-examples/example-groovy/build.gradle
```diff
@@ -2,8 +2,13 @@ plugins {
     id "groovy"
 }
 
+repositories {
+    mavenCentral()
+}
+
 dependencies {
     testImplementation project(":openapi")
+    testImplementation platform("io.micronaut:micronaut-bom:$micronautVersion")
```
ilopmar, line 11: I needed to add the repositories and the `platform` dependency (and the same in the java and kotlin projects) to be able to resolve the dependencies. Before the changes this wasn't necessary so I'm not sure if I'm missing something.

buildSrc/src/main/groovy/io.micronaut.build.internal.openapi-module.gradle
```diff
@@ -0,0 +1,9 @@
+plugins {
+    id "io.micronaut.build.internal.module"
+}
+
+tasks.withType(Test) {
```
melix, line 5: /!\ Use `tasks.withType(Test).configureEach`

docs-examples/example-groovy/build.gradle
melix: This is probably because before configuration was applied unconditionally on all subprojects, which isn't the case anymore.

melix (approved):
Minor comment regarding the missing `configureEach` which would trigger configuration of all test tasks even if they are not in the task graph.

PR #661, opened 2022-02-10
Multiple bug fixes with proper attribute inheritance

### LINKED ISSUE #659: Micronaut OpenAPI wrongly reuses schema description and default values like nullable (Leads to wrongly generated Swagger schema)
### Background

In one application we use the Micronaut OpenAPI generator for over 10 controllers and over 65 schema entities. Over time and after several upgrades, we realized that the entity/schema description in combination with references quite often changes (I didn't pay too much attention, so just commited the changed OpenAPI file into the frontend).

But today we ran into several problems that just generate a **wrong** OpenAPI file that breaks our Typescript frontend (We did some internal backend refactoring which disclosed these issues by accident).

Background: We have a system that sends emails to customers. We can read the current internal email configuration and change it. Everything is documented via OpenAPI.

Let's start with a simple enum that defines the email sending protocol + transport encryption:

```java
@Introspected
@Schema(description = "Schema that represents the possible email protocols + transport encryption for sending emails")
public enum EmailSendProtocolDto {
    SMTP,
    SMTP_SSL,
    SMTP_STARTTLS
}
```

And now a class that represents the current email configuration (+ add a GET endpoint):

```java
@Introspected
@Schema(description = "Schema that represents the current email settings")
public record ReadEmailSettingsDto(

        @Schema(description = "Flag that indicates whether the email sending is active or not. If set to false, all other values are null", required = true)
        Boolean active,

        @Schema(description = "Hostname or IP of the email server or null if email sending is disabled", required = true, nullable = true)
        String hostname,

        @Schema(description = "Port of the email server or null if email sending is disabled", required = true, nullable = true)
        Integer port,

        @Schema(description = "Protocol used for the connection or null if email sending is disabled", required = true, nullable = true)
        EmailSendProtocolDto protocol,

        @Schema(description = "Email username to login or null if email sending is disabled", required = true, nullable = true)
        String username,

        @Schema(description = "Plaintext password for the email user to login in or null if email sending is disabled", required = true, nullable = true)
        String plaintextPassword,

        @Schema(description = "Sender email address that is used to send emails or null if email sending is disabled", required = true, nullable = true)
        String senderEmail
) {
}
```

**Important**: It is possible to disable email sending, so the returned values, except active, in `ReadEmailSettingsDto` can be null (We don't return a 404, but null values. The active attribute specifies whether email sending is active or not).

This generates the following OpenAPI file:

```
openapi: 3.0.1
info:
  title: openapiproblems
  version: 0.0.1
paths:
  /api/email:
    get:
      tags:
      - Email
      summary: Get the email settings.
      description: Get the email settings.
      operationId: getEmailSettings
      parameters: []
      responses:
        "200":
          description: Email settings
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ReadEmailSettingsDto'
        "401":
          description: Not authenticated
        "500":
          description: Internal server error
components:
  schemas:
    EmailSendProtocolDto:
      required:
      - "true"
      type: string
      description: Protocol used for the connection or null if email sending is disabled
      nullable: true
      enum:
      - SMTP
      - SMTP_SSL
      - SMTP_STARTTLS
    ReadEmailSettingsDto:
      required:
      - active
      - hostname
      - plaintextPassword
      - port
      - protocol
      - senderEmail
      - username
      type: object
      properties:
        active:
          required:
          - "true"
          type: boolean
          description: "Flag that indicates whether the email sending is active or\
            \ not. If set to false, all other values are null"
        hostname:
          required:
          - "true"
          type: string
          description: Hostname or IP of the email server or null if email sending
            is disabled
          nullable: true
        port:
          required:
          - "true"
          type: integer
          description: Port of the email server or null if email sending is disabled
          format: int32
          nullable: true
        protocol:
          $ref: '#/components/schemas/EmailSendProtocolDto'
        username:
          required:
          - "true"
          type: string
          description: Email username to login or null if email sending is disabled
          nullable: true
        plaintextPassword:
          required:
          - "true"
          type: string
          description: Plaintext password for the email user to login in or null if
            email sending is disabled
          nullable: true
        senderEmail:
          required:
          - "true"
          type: string
          description: Sender email address that is used to send emails or null if
            email sending is disabled
          nullable: true
      description: Schema that represents the current email settings
```

## Current problems

We can identify several issues in this OpenAPI file:

## Problem 1: Wrong description

 The `EmailSendProtocolDto` entity is not described by it's own schema description (`Schema that represents the possible email protocols for sending emails`), but the first (or last?) attribute description of a referencing schema is taken. In our case this is `ReadEmailSettingsDto.protocol` with its description (`Protocol used for the connection or null if email sending is disabled`).

This becomes a problem as soon several schemas reference another schema like `EmailSendProtocolDto`. Then the first (last?) referencing attribute is taken. This can lead to documentation issues where one schema documentation leaks into all others that are referencing `EmailSendProtocolDto`.

### Problem 2: Inherited nullability (and probalbly required too)

Related to issue 1, the `nullable` value is also taken from the first referencing attribute (And probalbly other attributes like `required` too?). In our case the parameter `ReadEmailSettingsDto.procotol` can be null in case email sending is disabled. 

So the following wrong enum schema is generated (`nullable` should not be part of the enum itself, but the attribute of the other schema referencing it). In addition, the wrong description again:

```
components:
  schemas:
    EmailSendProtocolDto:
      required:
      - "true"
      type: string
      description: Protocol used for the connection or null if email sending is disabled <------------------------------ Wrongly inherited from the other schema referencing this schema
      nullable: true  <------------------------------ Wrongly inherited from the other schema referencing this schema
      enum:
      - SMTP
      - SMTP_SSL
      - SMTP_STARTTLS
```

In my example the `nullable = true` wins because it's the only example referencing `EmailSendProtocolDto`. But as soon as we add another schema referencing `EmailSendProtocolDto`, let's say `ReadEmailOutputLocationDto`, the schema for this new class is wrong too (+ add GET endpoint):

```java
@Introspected
@Schema(description = "Schema that represents an existing email output location")
public record ReadEmailOutputLocationDto(

        // Snipped a lot of attributes. Class doesn't make any sense for outstanders

        @Schema(description = "Protocol used for the connection", required = true)
        EmailSendProtocolDto protocol
) {
}
```

The enum is still wrong:

```
components:
  schemas:
    EmailSendProtocolDto:
      required:
      - "true"
      type: string
      description: Protocol used for the connection or null if email sending is disabled
      nullable: true
      enum:
      - SMTP
      - SMTP_SSL
      - SMTP_STARTTLS
```

From what I remember this also works the other way around: Set the parameter to `required=true` and `nullable=false` and it will not be possible to define a nullable attribute (I hope I got this observation correct)

### Nitpick 3: The required attributes are duplicated in the schema

The required attributes in a schema are duplicated (List of required attributes and for every attribute itself required)

### Expected Behaviour

* The plugin should use the schema description of the entity if available. If none is available, one can fallback to a referencing schema
* The plugin should use the required/nullable value of the entity if available, otherwise fallback to the default value
* Don't duplicate the required values

### Steps To Reproduce

1. Clone the repository https://github.com/swaechter/openapiproblems
2. Import the project into your IDE
3. Comment out all `@Controller` except the one in bug 10
4. Run the example and take a look at the generated OpenAPI file

### Environment Information

- Linux openjdk version "17.0.2" 2022-01-18
- Micronaut 3.2.7 with Micronaut OpenAPI 3.2.0

### Example Application

https://github.com/swaechter/openapiproblems (See bug 10)

### Version

3.2.7

n0tl3ss:
This code change fixes issue with inheritance of "description" and "nullable" attributes. The fields annotated with `@Schema` should always consider their "description" and "nullable" arguments. The "required" argument should be only displayed in the list of the required properties of the DTO.

Fix #659

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -799,6 +799,7 @@ abstract class AbstractOpenApiVisitor  {
                 final boolean required = hasElementSchemaRequired(element).orElseGet(() -> isElementNotNullable(element, classElement));
                 propertySchema = bindSchemaForElement(context, element, elementType, propertySchema);
                 String propertyName = resolvePropertyName(element, classElement, propertySchema);
+                propertySchema.setRequired(null);
                 addProperty(parentSchema, propertyName, propertySchema, required);
             }
         }
@@ -885,6 +886,12 @@ abstract class AbstractOpenApiVisitor  {
         Schema originalSchema = schemaToBind;
         if (originalSchema.get$ref() != null) {
             schemaToBind = new Schema();
+            if (schemaAnn != null) {
+                Optional<String> schemaDescription = schemaAnn.get("description", String.class);
+                if (schemaDescription.isPresent()) {
+                    schemaToBind.setDescription(schemaDescription.get());
+                }
+            }
         }
         if (originalSchema.get$ref() == null && schemaAnn != null) {
             // Apply @Schema annotation only if not $ref since for $ref schemas
@@ -960,7 +967,7 @@ abstract class AbstractOpenApiVisitor  {
         }
         // @Schema annotation takes priority over nullability annotations
         Boolean isSchemaNullable = element.booleanValue(io.swagger.v3.oas.annotations.media.Schema.class, "nullable").orElse(null);
-        if ((isSchemaNullable == null && element.isNullable()) || (Boolean.TRUE.equals(isSchemaNullable) && element.isNullable())) {
+        if ((isSchemaNullable == null && element.isNullable()) || Boolean.TRUE.equals(isSchemaNullable)) {
             schemaToBind.setNullable(true);
         }
         final String defaultJacksonValue = element.stringValue(JsonProperty.class, "defaultValue").orElse(null);
@@ -1244,6 +1251,17 @@ abstract class AbstractOpenApiVisitor  {
                 inProgressSchemas.add(schemaName);
                 try {
                     schema = readSchema(schemaValue, openAPI, context, type, mediaTypes);
+                    AnnotationValue<io.swagger.v3.oas.annotations.media.Schema> typeSchema = type.getDeclaredAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
+                    if (typeSchema != null) {
+                        Schema originalTypeSchema = readSchema(typeSchema, openAPI, context, type, mediaTypes);
+
+                        if (originalTypeSchema.getDescription() != null && !originalTypeSchema.getDescription().isEmpty()) {
+                            schema.setDescription(originalTypeSchema.getDescription());
+                        }
+                        schema.setNullable(originalTypeSchema.getNullable());
+                        schema.setRequired(originalTypeSchema.getRequired());
+                    }
+
                     if (schema != null) {
                         schema.setName(schemaName);
                         schemas.put(schemaName, schema);
--- a/openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiRecursionSpec.groovy
+++ b/openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiRecursionSpec.groovy
@@ -175,8 +175,10 @@ class MyBean {}
         expect:
             Schema testImpl1 = schemas.get("TestImpl1")
             Schema woopsieRef = testImpl1.getProperties()["woopsie"]
-            woopsieRef
-            woopsieRef.$ref == "#/components/schemas/woopsie-id"
+
+            woopsieRef instanceof ComposedSchema
+            ((ComposedSchema) woopsieRef).allOf[0].$ref == "#/components/schemas/woopsie-id"
+            ((ComposedSchema) woopsieRef).allOf[1].description == "woopsie doopsie"
             Schema woopsie = schemas.get("woopsie-id")
             woopsie.description == "woopsie doopsie"
     }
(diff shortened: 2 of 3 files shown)


graemerocher (approved):

swaechter:
@n0tl3ss I just tested your fix with Micronaut 3.3.4 and Micronaut OpenAPI 4.0.1 Snapshot. A lot of our issues were fixed. A big thank you!

I found one issue regarding the nullability of a type. The new generated code looks like this (Attributes like protocol can be null when email sending is disabled):

```yaml
    ReadEmailOutputLocationDto:
      required:
      - protocol
      type: object
      properties:
        protocol:
          allOf:
          - $ref: '#/components/schemas/EmailSendProtocolDto'
          - description: Protocol used for the connection
      description: Schema that represents an existing email output location
    ReadEmailSettingsDto:
      required:
      - active
      - hostname
      - plaintextPassword
      - port
      - protocol
      - senderEmail
      - username
      type: object
      properties:
        active:
          type: boolean
          description: "Flag that indicates whether the email sending is active or\
            \ not. If set to false, all other values are null"
        hostname:
          type: string
          description: Hostname or IP of the email server or null if email sending
            is disabled
          nullable: true
        port:
          type: integer
          description: Port of the email server or null if email sending is disabled
          format: int32
          nullable: true
        protocol:
          allOf:
          - $ref: '#/components/schemas/EmailSendProtocolDto'
          - description: Protocol used for the connection or null if email sending
              is disabled
            nullable: true
        username:
          type: string
          description: Email username to login or null if email sending is disabled
          nullable: true
        plaintextPassword:
          type: string
          description: Plaintext password for the email user to login in or null if
            email sending is disabled
          nullable: true
        senderEmail:
          type: string
          description: Sender email address that is used to send emails or null if
            email sending is disabled
          nullable: true
      description: Schema that represents the current email settings
```

Problematic is the location of the nullable:

```yaml
    ReadEmailSettingsDto:
      required:
        SNIPPED
      type: object
      properties:
        SNIPPED OTHER PROPERTIES
        protocol:
          allOf:
          - $ref: '#/components/schemas/EmailSendProtocolDto'
          - description: Protocol used for the connection or null if email sending
              is disabled
            nullable: true <------------------------------------- Should not be down here
```

According to https://stackoverflow.com/questions/40920441/how-to-specify-a-property-can-be-null-or-a-reference-with-swagger the null should be at before the allOf, so like this:

```yaml
    ReadEmailSettingsDto:
      required:
        SNIPPED
      type: object
      properties:
        SNIPPED OTHER PROPERTIES
        protocol:
          nullable: true <------------------------------------- Moved up here
          allOf:
          - $ref: '#/components/schemas/EmailSendProtocolDto'
          - description: Protocol used for the connection or null if email sending
              is disabled
```

Moving all nullable: true in our code fixes the OpenAPI generation.

PR #666, opened 2022-02-22
update server-context documentation

StefanKreidel:
As suggested in micronaut-projects/micronaut-core#6905 I have updated the documentation to include a way to handle dynamic context-paths at runtime.
This is very useful if the same build artifact is intended to run behind multiple reverse proxies (e.g. on different stages) which serve the application on different context-paths.

The idea is to parse the context-path from the request headers. This assumes that the reverse proxy is configured to set the header accordingly.

Unfortunately, I could not test if the docs still compile correctly. If there is an easy way to test that, I'm happy to do so. Just let me know.


StefanKreidel:
Any of the maintainers willing to have a look at this?

sdelamo:
Thanks for the contribution @StefanKreidel 

> Unfortunately, I could not test if the docs still compile correctly. If there is an easy way to test that, I'm happy to do so. Just let me know.

We could move the code samples to actual code within the `docs-examples` folder.

PR #729, opened 2022-05-24
Fixed controller search in inner classes. Added lost property naming strategies.

### LINKED ISSUE #696: Controller annotation not found when defined as inner class
### Expected Behavior

Given the following definition:

```java
public sealed interface QuoteApi {

    List<Quote> listQuotes();

    non-sealed interface QuoteClient extends QuoteApi {
        @Get
        @Override
        List<Quote> listQuotes();
    }

    @Controller("/quotes")
    final class QuoteController implements QuoteApi {
        @Get("/")
        @Override
        public List<Quote> listQuotes() {
            ...
        }
    }
}
```

The `QuotesController` should trigger generation of OpenAPI metadata.

### Actual Behaviour

Metadata isn't generated. Moving the `QuoteController` as a top-level class fixes the problem.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.4.1

altro3:
Fixed #696


graemerocher:
Thanks for the contribution!

PR #730, opened 2022-05-24
Added support for enum values with JsonProperty annotation

### LINKED ISSUE #471: JsonProperty not getting picked up by yaml generation for Enum
### Task List

- [ X ] Steps to reproduce provided
- [ N/A] Stacktrace (if present) provided
- [ N/A] Example that reproduces the problem uploaded to Github
- [ X ] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create Java Enum Class with JsonProperty
```

import com.fasterxml.jackson.annotation.JsonProperty;

public enum CurrencyPair2 {
    @JsonProperty("CAD-USD") CADUSD("CAD", "USD");
    private final String baseCurrency;
    private final String quoteCurrency;

    CurrencyPair2(String baseCurrency, String quoteCurrency) {
        this.baseCurrency = baseCurrency;
        this.quoteCurrency = quoteCurrency;
    }

    public String getBaseCurrency() {
        return baseCurrency;
    }

    public String getQuoteCurrency() {
        return quoteCurrency;
    }

}

```
2. Reference class in response object/request params
3. Build OpenApi:
```

annotationProcessor("io.micronaut.openapi:micronaut-openapi")
...
implementation("io.swagger.core.v3:swagger-annotations:2.1.6")
...

```
tasks.withType(JavaCompile) {
    options.fork = true
    options.forkOptions.jvmArgs << '-Dmicronaut.openapi.views.spec=rapidoc.enabled=true,swagger-ui.enabled=true,swagger-ui.theme=flattop'
}

4. Check Swagger output

### Expected Behaviour

Enum values should be CAD-USD and serialization should convert appropriately
```
CurrencyPair2:
      type: string
      enum:
      - CAD-USD

```

### Actual Behaviour
Serialization converts on ingress and egress appropriately but enum values are CADUSD in the swagger schema
```
CurrencyPair2:
      type: string
      enum:
      - CADUSD

```

Tell us what happens instead

### Environment Information

- **Operating System**: MavOS Mojave version 10.14.6 
- **Micronaut Version:** 1.3.4
- **JDK Version:** 
- OPENAPI_VERSION=2.0.10

### Example Application

- TODO: link to github repository with example that reproduces the issue

altro3:
Fixed #471

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+
 import io.micronaut.core.annotation.AnnotationClassValue;
 import io.micronaut.core.annotation.AnnotationValue;
 import io.micronaut.core.annotation.Nullable;
@@ -89,6 +90,7 @@ import io.swagger.v3.oas.models.security.SecurityRequirement;
 import io.swagger.v3.oas.models.security.SecurityScheme;
 import org.reactivestreams.Publisher;
 
+import javax.lang.model.element.ElementKind;
 import javax.validation.constraints.DecimalMax;
 import javax.validation.constraints.DecimalMin;
 import javax.validation.constraints.Email;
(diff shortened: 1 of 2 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -1368,6 +1370,22 @@ private Schema getSchemaDefinition(
...
+                JsonProperty jsonproperty = element.getAnnotation(JsonProperty.class);
+                if (jsonproperty != null && StringUtils.hasText(jsonproperty.value())) {
+                    enumValues.add(jsonproperty.value());
+                } else {
+                    enumValues.add(element.getSimpleName().toString());
+                }
+            }
+        }
+        return enumValues;
+    }
```
graemerocher, line 1387: This change will only work on Java, we prefer to not have direct dependencies on the `javax.lang.model` api. Can you alter this to use `EnumElement` and the `values()` method is already offers. This possibly will require a change in core.

altro3: @graemerocher Hmm... if this method cannot be used - then yes, you will have to modify the kernel. Because now it is possible to get only a list of values, without annotations (Ivan described the problem in the issue). In general, need to add a method to `EnumElement` that will return a list of values and their annotations

altro3:
Fixed here: https://github.com/micronaut-projects/micronaut-openapi/pull/791

PR #719, opened 2022-05-20
Fix problem with @Encoding. Uparde libs to actual versions.

altro3:
Fixed #718

PR #732, opened 2022-05-26
Line endings

timyates:
This PR https://github.com/micronaut-projects/micronaut-openapi/pull/719

Set all line endings to LF, even though the gradlew.bat file was CRLF, so when you clone the repo it immediately wants to change the line endings on the file

This PR uses the .gitattributes from the template project and sets the line endings of gradlew.bat correctly


timyates:
Checked with

```
git clone -b line-endings --single-branch https://github.com/micronaut-projects/micronaut-openapi.git openapi-lines
```

And you get a project where the bat file is CRLF, and the repo doesn't start off with changed files

sdelamo (approved):

PR #739, opened 2022-06-05
Fixed encoding style parameter. Fix lost schema ref

altro3:
See test. Left file generated by official swagger plugin, right - by micronaut-openapi

изображение


graemerocher:
Thanks for the contribution!

PR #743, opened 2022-06-12
Added support ExternalDocumentation annotation

altro3:
Added support ExternalDocumentation annotation, improved test for complex  Content block


graemerocher:
Thanks for the contribution!

PR #745, opened 2022-06-12
Various improvements to schema binding

altro3:
- kotlin 1.7.0
 - Improved test for complex OpenAPIDefinition
 - Fixed lost default value in schema (on class level)
 - Fixed lost default, example, deprecated, accessMode properties, when use Schema on property level
 - Fixed rename property when use Schema on property level


openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -990,6 +997,39 @@ protected Schema bindSchemaForElement(VisitorContext context, Element element, C
...
+                        schemaToBind.setReadOnly(true);
+                        schemaToBind.setWriteOnly(null);
+                    } else if (schemaAccessMode.get() == io.swagger.v3.oas.annotations.media.Schema.AccessMode.WRITE_ONLY) {
+                        schemaToBind.setReadOnly(false);
+                        schemaToBind.setWriteOnly(null);
+                    } else if (schemaAccessMode.get() == io.swagger.v3.oas.annotations.media.Schema.AccessMode.READ_WRITE) {
+                        schemaToBind.setReadOnly(null);
+                        schemaToBind.setWriteOnly(null);
+                    }
+                }
```
graemerocher, line 1032: Not critical but this code can be written more cleanly with something like:

```java
schemaAnn.enumValue("accessMode", io.swagger.v3.oas.annotations.media.Schema.AccessMode.class)
   .ifPresent(accessMode -> {
          switch(accessMode) {
               case READ_ONLY:
                        schemaToBind.setReadOnly(true);
                         schemaToBind.setWriteOnly(null);
                break;
               case WRITE_ONLY:
                        schemaToBind.setReadOnly(false);
                         schemaToBind.setWriteOnly(null);
                break;
          }
   });
```

Apart from that is setting write only to null correct?

altro3: @graemerocher I don't now, soory. This nlock copypasted from swagger lib

altro3: @graemerocher Your proposal is not so simple, because would require making the schemaToBind variable finalized. In this case, I think that this block is better left as is.

изображение

altro3:
@graemerocher Hi! Could anyone review my pull requests? Because I want to continue improve this module and now it's already **7 PR** open from me and need to merge them so that there are fewer problems with conflicts

graemerocher:
@altro3 was away last week, apologies for the delay

PR #749, opened 2022-06-19
Support JSON format

### LINKED ISSUE #102: Ability to generate spec in json instead of yaml
Hi,

I would like to know if there is any OOTB or an easy extension point to generate the OpenAPI spec in json format instead of Yaml. 

Thanks.

altro3:
Add system property (micronaut.openapi.json.format) to change format of output file - JSOn or YAML.

Fixes #102


graemerocher:
Thanks for the contribution!

PR #772, opened 2022-07-16
Fixes schema processing

altro3:
- Gradle 7.5
 - Ignore default values in Schema annotation.
 - Fix lost anyOf/allOf/oneOf/not schema properties.
 - Fix generating empty parameters list.
 - Fix invalid reading Response content schema.
 - Move some methods and field to Utils class.
 - Move some methods and field to ConvertUtils class.
 - Fix problems for all schemas (in Parameter, Header, Content, etc)


altro3:
@graemerocher hi! I think need release these changes to 4.4.1, because I made mistake and forgot about default values in Schema annotation. Look to openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java (https://github.com/micronaut-projects/micronaut-openapi/pull/772/files#diff-70483a9ff868dce418c3a545df1929d1c33a917d15e9bcc650b3e5185166b909): I processed default values as real values and it was mistake.

I fixed it and also I fixed global problem with schemas (some values were lost every time). Problem was in convertion from JsonNode to openApi objects. I don't know why, but standard swagger ObjectMapper convert the values to JsonNode correctly, but loses some of the values when converting them back. Therefore, need to use another mapper for correct reverse conversion.

graemerocher:
@altro3 ok I will do a release soon. Thanks

PR #782, opened 2022-07-29
Expose libraries required for Groovy projects

### LINKED ISSUE #781: Compilation failure in Groovy projects since Micronaut OpenAPI 4.4.0
@altro3 FYI, #761 has introduced a regression in Groovy projects. They won't compile because both `com.github.chhorz:javadoc-parser` and `com.vladsch.flexmark:flexmark-html2md-converter` need to be in the classpath (for Groovy projects):

https://github.com/micronaut-projects/micronaut-starter/runs/7563422170?check_suite_focus=true#step:5:3218

I wonder why the test suite here didn't catch this problem. Will investigate

alvarosanchez:
Groovy projects need `io.micronaut.openapi:micronaut-openapi` in the
compile classpath for the AST transformations to work (unlike Java/Kotlin,
which have it on the annotation processor scope only).

Fixes #781

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,7 +11,7 @@ title=OpenAPI/Swagger Support
 projectDesc=Configuration to integrate Micronaut and OpenAPI/Swagger
 projectUrl=https://micronaut.io
 githubSlug=micronaut-projects/micronaut-openapi
-developers=Puneet Behl,√Ålvaro S√°nchez-Mariscal,Iv√°n L√≥pez
+developers=Puneet Behl,¡lvaro S·nchez-Mariscal,Iv·n LÛpez
 
 githubCoreBranch=3.6.x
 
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -11,7 +11,7 @@ title=OpenAPI/Swagger Support
 projectDesc=Configuration to integrate Micronaut and OpenAPI/Swagger
 projectUrl=https://micronaut.io
 githubSlug=micronaut-projects/micronaut-openapi
-developers=Puneet Behl,Álvaro Sánchez-Mariscal,Iván López
+developers=Puneet Behl,�lvaro S�nchez-Mariscal,Iv�n L�pez
```
alvarosanchez, line 14: This was being displayed as:

image

🤷

alvarosanchez: In fact, it is wrong in the published POMs: https://search.maven.org/artifact/io.micronaut.openapi/micronaut-openapi/4.4.2/jar

sdelamo: I blame you for having a tilde in your name

sdelamo (approved):

PR #790, opened 2022-08-07
Resource localization for UI

### LINKED ISSUE #779: Swagger UI is using external resources
### Expected Behavior

When using `swagger-ui.enabled=true` swagger should be entirely self-contained, not depending on any external resources.

### Actual Behaviour

The UI requires external resources, see screenshot below:
image

This is an issue for multiple reasons:
1. It leaks internal urls to an external CDN
2. unpkg.com is very slow (for whatever reason) causing up to 40 seconds to load

### Steps To Reproduce

1. Enable swagger ui: `swagger-ui.enabled=true` 
2. Open swagger-ui and look at the requests

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.5.3

altro3:
Now all the necessary files for the correct operation of the UI (redoc, rapidoc, rapipdf, swagger-ui) are fully localized. Added the ability to set custom paths to the required resources (property `*.js.url`), if, for example, someone wants to use loading not from a local server, but from external resources

Fixes #779


altro3:
@graemerocher Hi! Could you exclude files for UI for checkstyle?
изображение

graemerocher:
Supressions can be configured with https://github.com/micronaut-projects/micronaut-openapi/blob/master/config/checkstyle/suppressions.xml

altro3:
@graemerocher Hm... i tried many ways to add suppresion for this check, but can't disable it :-(. For example, I tried to exclude `rapidoc-min.js` file for this checking. My variants:
 - rapidoc-min.js
 - rapidoc\-min\.js
 - .*rapidoc-min.*
 - .*rapidoc\-min.*
 - .*rapidoc.*
 - .*

None of these options work. I'm leaning towards that this file is simply ignored no matter what is specified there.

graemerocher:
Seems it is the nohttp plugin which prevents the use of http:// which is different from the checkstyle plugin

graemerocher:
Think a file similar to https://github.com/swapnard/java/blob/fa1eb544c159076389580134540988030b4724ea/src/nohttp/checkstyle.xml has to be defined

altro3:
@graemerocher Hi! I tried to use these settings , but gradle ignore my changes. You can see my changes in pull request:
изображение
But absolutely nothing has changed.

I dont know how fix it :-(

altro3:
@graemerocher Again tried to configure this spring nohttp plugin - it's not condigurable :-(. I tried to set exclusions in supressions file:
```
<suppressions>
    <suppress files="rapidoc-min.js" checks=".*"/>
    <suppress files="rapipdf-min.js" checks=".*"/>
    <suppress files="redoc.standalone.js" checks=".*"/>
    <suppress files="swagger-ui-bundle.js" checks=".*"/>
    <suppress files="swagger-ui-standalone-preset.js" checks=".*"/>
</suppressions>
```
Tried to cofigure this plugin in checkstyle.xml:
```
    <module name="io.spring.nohttp.checkstyle.check.NoHttpCheck">
        <property name="allowlistFileName" value="${config_loc}/allowlist.lines"/>
    </module>
```
But this spring plugin simply ignores absolutely all settings. How to be? Maybe nned to disable it from this project?

graemerocher:
perhaps just disable `checkstyleNohttp` task for this project

altro3:
@graemerocher Hi! yes, I agree. I tried to do it, but couldn't. Need help :-(

altro3:
@graemerocher Hi! Could anyone help? Can't disable this rule because it is built into the spring library and connected using your plugin. As I said - I tried different ways, but all my changes seem to be simply ignored by the checkstyle plugin :-(

graemerocher:
going to have a look

vtyrog:
please add info about configuration `swagger-ui.js.url` in documentation

PR #791, opened 2022-08-07
Bug fixes. Add support enum annotations

altro3:
- Added support for enum values with JsonProperty annotation (fix #471).
 - Added support JsonAnySetter annotation (fix #138)
 - Fix wrong logic for property name when used Schema and JsonProperty annotations (fix #787)
 - Fixed #134
 - Fix incorrect schema when set type for property level annotation (fixed #792)
 - Fix field javadoc properties parsing.
 - Fix invalid defaultValue and allowValues types (have always been String)
 - Fix `example` field with null value (always added `example: null` to yaml file)
 - Removed unnecessary `required: false` field in final yaml file
 - Upgrade libs:

Micronaut 3.6.1
Gradle 7.5.1
Groovy 3.0.12

PR #794, opened 2022-08-21
Environments support

### LINKED ISSUE #166: Not resolving Micronaut config values in Controller annotation
Hi,

I have some config in my `@Controller` annotation, like this:

```
@Controller("/v${applications.api.version}/applications")
```

And my OpenAPI is being generated like this:

```
  /v${applications.api.version}/applications:
    get:
```

I believe that correct would be:

```
  /v1/applications:
    get:
```
Because I've stetted `${applications.api.version}` to 1:

Micronaut works fine, I'm able to resolve the controller `/v1/applications`.

Thanks a lot and great project!

### LINKED ISSUE #759: Micronaut OpenApi: Expand also placeholder not only in openapi.properties
### Feature description
**Current Behavior**
@OpenAPIDefinition only accepts constant values and placeholder from openapi.properties

**Expected Behavior**

- [ ]  Can we also expand the value for micronaut.server.context-path which in application.yml 
- [ ] Can we also expand or way to use the build-info.properties, build.version for OpenAPIDefinition

**Example**

_For Task 1_
```
@OpenAPIDefinition(
    info = @Info(servers = {@Server(url ="${micronaut.server.context-path}")	
    }
```

We intend to override the server url to context-path due to CORS issue upon invoking request with swagger-ui it was on HTTP while server host is at HTTPS

_For Task 2_
```
@OpenAPIDefinition(
    info = @Info(version  ="${build.version}")
 }
```

 Since the version in openapi definition is mostly same with the current version of the application service

### LINKED ISSUE #766: Placeholder expansion cannot be overridden by system properties.
### Expected Behavior

Based on these lines from the documentation:

> It is possible to tweak the OpenAPI processing with system properties or with a properties file. Options specified with system properties have priority over those defined in the openapi.properties file

> Properties prefixed with micronaut.openapi.expand will be expanded at compile time, for instance with:

it seems like it should be possible to supply `micronaut.openapi.expand.x` as a system property to inject values for placeholder expansion. 

### Actual Behaviour

Expandable placeholders are read from openapi.properties [here](https://github.com/micronaut-projects/micronaut-openapi/blob/29fb37e1d5b0d59401050e92ea70e397f6960dda/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java#L503), and therefore only values supplied in openapi.properties can be used to expand placeholders. 

Would it be possible to expand placeholders including overrides from system properties? Or failing that, at least update the documentation to reflect the fact that only values directly supplied from openapi.properties will be expanded?

Thank you!

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

latest

altro3:
Finally added support standard properties application.yml with environments

Fixes #766 
Fixes #759 
Fixes #166

Merge it after https://github.com/micronaut-projects/micronaut-openapi/pull/791

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.openapi.visitor;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import io.micronaut.context.ApplicationContextConfiguration;
+import io.micronaut.context.env.ActiveEnvironment;
+import io.micronaut.context.env.CachedEnvironment;
+import io.micronaut.context.env.DefaultEnvironment;
+import io.micronaut.context.env.Environment;
+import io.micronaut.context.env.PropertiesPropertySourceLoader;
+import io.micronaut.context.env.PropertySource;
+import io.micronaut.context.env.PropertySourceLoader;
+import io.micronaut.context.exceptions.ConfigurationException;
+import io.micronaut.core.io.ResourceLoader;
+import io.micronaut.core.io.file.DefaultFileSystemResourceLoader;
+import io.micronaut.core.order.OrderUtil;
+import io.micronaut.core.util.StringUtils;
+import io.micronaut.inject.visitor.VisitorContext;
+
+import static io.micronaut.openapi.visitor.OpenApiApplicationVisitor.MICRONAUT_CONFIG_FILE_LOCATIONS;
+
+/**
+ * Specific environment for annotation processing level. Solve problem with access to resources
+ * from project classpath.
+ *
+ * @since 4.5.0
+ */
+public class AnnProcessorEnvironment extends DefaultEnvironment {
+
+    private final List<String> annotationProcessingConfigLocations;
+    private String projectResourcesPath;
+    private String projectDir = StringUtils.EMPTY_STRING;
+
+    /**
+     * Construct a new environment for the given configuration.
+     *
+     * @param configuration The configuration
+     * @param context visitor context
+     */
+    public AnnProcessorEnvironment(ApplicationContextConfiguration configuration, VisitorContext context) {
+        super(configuration);
+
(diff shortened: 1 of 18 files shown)


graemerocher:
@altro3 could you rebase? Thanks

altro3:
@graemerocher yes, sure. But, please, don't merge this PR yet. I'm not sure that everything is working as intended at the moment. In Java, there are restrictions on the access of the annotation processor to the source code. Those. I'm not sure if this solution will work, I need to additionally test the library when it is included as an annotation processor. I'll try to test it over the weekend

altro3:
Phew, I had to complicate the solution a bit due to java limitations, but I implemented the ability to use standard configurations for micronaut. I had to add the project keyword to identify the path to the yaml files that are in the project sources, because. when starting annotation processing, java does not provide access to the project files directly, that is, it does not load classes into the classpath and does not load resources.

Pull request can be merged

altro3:
@graemerocher Phew, I had to complicate the solution a bit due to java limitations, but I implemented the ability to use standard configurations for micronaut. I had to add the project keyword to identify the path to the yaml files that are in the project sources, because. when starting annotation processing, java does not provide access to the project files directly, that is, it does not load classes into the classpath and does not load resources.

Pull request can be merged

openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Construct a new environment for the given configuration.
+     *
+     * @param configuration The configuration
+     * @param context visitor context
+     */
+    public AnnProcessorEnvironment(ApplicationContextConfiguration configuration, VisitorContext context) {
+        super(configuration);
+
+        Path projectPath = context.getProjectDir().orElse(Paths.get(System.getProperty("user.dir")));
```
graemerocher, line 67: don't think we should be using the user directory like this. Better to just use `null` and do nothing

openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
```diff
@@ -0,0 +1,180 @@
...
+        projectDir = projectPath.toString().replaceAll("\\\\", "/");
```
graemerocher, line 68: this seems like a hack, why can't you just use `path.toUri()`?

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
altro3: @graemerocher The problem is that when the tests are run, their directory is empty, because the processor is started for the generated classes. I agree that user.dir is only required for tests. Fixed

openapi/src/main/java/io/micronaut/openapi/visitor/AnnProcessorEnvironment.java
altro3: @graemerocher Don't agree with you. This a stnadrad fix for windows. Why didn't I use `toUri` ? Let's see to example. My directory is `D:\Work\my-project` in windows and `/home/my/work/my-project`. I need to set this path as `file:D:\Work\my-project` for wind and `file:/home/my/work/my-project`.

If I use `path.toUri().toString()` than I have these paths: `file:///D:/Work/my-project/` for Windows and  `file:///D:/Work/my-project/` and `file:///home/my/work/my-project/`.

The problem is that in order to correctly display the correct path, I need to trim file and 3 slashes for Windows, and file and 2 slashes for Linux. In general, there are more problems than benefits from such a solution. My solution is simple and clear

PR #803, opened 2022-09-13
Fixed incorrect logic for copy UI resources from micronaut-openapi jar file.

altro3:
Fixed incorrect logic for copy UI resources from micronaut-openapi jar file.

Add checks for empty javadocs.


altro3:
@graemerocher Hi! Could you merge this chages before release? There are very important fixes

PR #831, opened 2022-10-23
Move to micronaut 4.0.0

altro3:
Move to micronaut 4.0.0.


wetted:
@altro3 see #834 

It has some issues with broken tests, but @sdelamo will end up merging to the 5.0.x branch, and then end up merging that to master to make it the tracking branch for Micronaut 4.0.x. Depending on timing for the open PRs in the description you need merged, this will target either the 5.0.x branch or master. Hope that all makes sense.

altro3:
@sdelamo @graemerocher Could ypu merge it to master? In this PR I fixed all the problems that occurred with macronaut 4. Changed the logic so that everything works with macronaut 4.0.0

graemerocher:
the build is failing?

altro3:
@graemerocher yes, need help. After last changes cna't set dependency to micronuat-security. Could you tell me how I can set this dependecy now (it works before last changes in micronaut-core):
```
    testImplementation(mn.micronaut.security)
```

graemerocher:
you need to add a dependency on the Micronaut Security BOM

altro3:
@graemerocher Sorry, don't understand. I need to set version of mivronaut security in libsversions catlogue also? Or how I can get micronaut-security version now? Can i connect micronaur-security version like this:
```groovy
micronautBuild {
    importMicronautCatalog()
}
```
?
In general, I'm confused about how to work with micronaut versions now. I would like not to duplicate a large number of versions in different projects, but to use some kind of single set of versions

graemerocher:
Yes similar to https://github.com/micronaut-projects/micronaut-micrometer/blob/4f3a0560eee886d487a86dcb6c76297779d10e86/settings.gradle#L41

graemerocher:
The reason is we split out the BOM and now modules can't depend directly on the BOM. See https://github.com/micronaut-projects/micronaut-platform

graemerocher:
Note you have to use a custom prefix when importing the BOM, so you declare:

```groovy
 importMicronautCatalog("micronaut-security")
```

Then use:

```
testImplementation mnSecurity.security.core
```

altro3:
Hmm.. 
```
Version catalog doesn't contain a library with alias: micronaut-security
```

изображение

wetted:
You have to define it in the module version catalog. See: https://github.com/micronaut-projects/micronaut-micrometer/blob/4f3a0560eee886d487a86dcb6c76297779d10e86/gradle/libs.versions.toml#L20

altro3:
Does `micronaut-security` version catalog exist?

altro3:
@wetted Is there a solution where I don't need to specify the security version? Just then it will be necessary to update not only the micronaut version, but also the security version

graemerocher:
there is no solution for that no, but renovate will send PRs to update this.

altro3:
@graemerocher @wetted Ok, I changed dependencies for versions catalogs. What problems I see now:

1. `testRuntimeOnly(mnTest.junit.jupiter.engine)` - it doesn't work now. It's not clear how to get the junit version and micronaut-test
2. Can't get `jackson-datatype-jsr310` and `jackson-dataformat-yaml` from micornaut-bom, фlthough jackson-bom is there
3. `jakarta-validation-api` - micronaut-validation still doesn't support jakarta.validation-api and I can't get a version of this library from there
4. I don’t understand why Kotlin versions are not marked managed in the micronaut-kotlin module. Because of this, I can't use the versions for kotlin-stdlib-jdk8, kotlin-reflect, kotlin-jvm, kotlin-kapt and kotlin-allopen

Of course, I can understand something wrong, but it seems to me that the less duplication of versions, the better.

altro3:
@graemerocher @wetted Ok, fixed 1 and 2 points. All tests passed now, but a strange problem appeared in the Kotlin example, for some reason the junit-jupiter version is not picked up. At the same time, exactly the same code works fine in the java example. Can you help figure it out?

graemerocher:
the kotlin version is missing `testImplementation(mnTest.micronaut.test.junit5)`

altro3:
@graemerocher Thanks! fixed

PR #840, opened 2022-11-12
Add ability to customize classes schemas in properties.

### LINKED ISSUE #654: Convert types (for example ObjectId)
### Feature description

Not sure if you support this already, but I couldn't find much in the documentation.

I'd like to map Mongo ObjectIds inside a POJO to a string, instead of a complex object.

Is a ModelConverter the way to go?
https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-Configuration#configuration-properties
If so, how can I configure it? Could you provide an example?

If possible, I'd like to use my existing Jackson JsonDeserializer or the Micronaut TypeConverter.

I already tried to use the Schema-annotation of a field of a Kotlin data class, but it seemed to be ignored.
Also, I'd rather configure it once for the project and not on every field of this type.

https://stackoverflow.com/questions/58375613/swagger-shows-mongo-objectid-as-complex-json-instead-of-string
https://github.com/swagger-api/swagger-core/issues/720
https://github.com/swagger-api/swagger-core/issues/3718

Thanks in advance
Andreas

altro3:
Now you can set custom classes to create different open api schemas for selected classes:
```yaml
micronaut:
  openapi:
    schema:
      org.somepackage.MyComplexType: java.lang.String
      org.somepackage.MyComplexType2: java.lang.Integer
```
or by system properties:
```
-Dmicronaut.opemapi.schema.org.somepackage.MyComplexType=java.lang.String -Dmicronaut.opemapi.schema.org.somepackage.MyComplexType2=java.lang.Integer
```
or by `openapi.properties`
```properties
micronaut.opemapi.schema.org.somepackage.MyComplexType=java.lang.String
micronaut.opemapi.schema.org.somepackage.MyComplexType2=java.lang.Integer
```

Also It can be used for replace classes schema with generics, for example, if you use jaxb generated classes and have custom serializer for JAXBElement class (see tests). And you can set cutom schemas for different type args. For example if you have this classes structure:
```java
package test.mypackage;

class MyDto {

    public JAXBElement<? extends XmlElement> xmlElement;
    public JAXBElement<? extends XmlElement2> xmlElement2;
    public JAXBElement<? extends XmlElement3> xmlElement3;
}

package test.mypackage;

class XmlElement {
    public String propStr;
}

package test.mypackage;

class XmlElement2 {
    public String propStr2;
}

package test.mypackage;

class XmlElement3 {
    public String propStr3;
}
```

You can customize classes structure for openapi schema:
```java
package io.micronaut.openapi;

// if you want to use generic from fields with type JAXBElement<T>
class MyJaxbElement<T> {
    public String type;
    public T value;
}

package io.micronaut.openapi;

class MyJaxbElement2 {
    public String type;
    public List<String> values;
}

package io.micronaut.openapi;

class MyJaxbElement3 {
    public String type;
    public String value;
}
```
And set openapi properties to map classes to custom openapi schema classes:
```yaml
micronaut:
  openapi:
    schema:
      io.micronaut.openapi.JAXBElement: io.micronaut.openapi.MyJaxbElement
      io.micronaut.openapi.JAXBElement<test.mypackage.XmlElement2>: io.micronaut.openapi.MyJaxbElement2
      io.micronaut.openapi.JAXBElement<test.mypackage.XmlElement3>: io.micronaut.openapi.MyJaxbElement3
```

Fixed #654

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,8 +1,8 @@
-projectVersion=4.6.1-SNAPSHOT
+projectVersion=4.7.0-SNAPSHOT
 projectGroup=io.micronaut.openapi
 
 micronautDocsVersion=2.0.0
-micronautVersion=3.7.2
+micronautVersion=3.7.3
 micronautTestVersion=3.7.0
 groovyVersion=3.0.13
 spockVersion=2.3-groovy-3.0
@@ -13,7 +13,7 @@ projectUrl=https://micronaut.io
 githubSlug=micronaut-projects/micronaut-openapi
 developers=Puneet Behl,Álvaro Sánchez-Mariscal,Iván López
 
-githubCoreBranch=3.7.x
+githubCoreBranch=3.8.x
 
 bomProperty=micronautOpenapiVersion
 bomProperties=swaggerVersion
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -10,7 +10,7 @@ managed-slf4j = "1.7.36"
 # Versions beyond 0.62.2 require Java 11
 managed-html2md-converter = "0.62.2"
 
-kotlin = "1.7.10"
+kotlin = "1.7.21"
 
 [libraries]
 # Duplicated to keep catalog compatibility with 3.4.x.  Can be removed for 4.0.0
--- a/gradle/license.gradle
+++ b/gradle/license.gradle
@@ -9,7 +9,7 @@ license {
         java   = 'SLASHSTAR_STYLE'
         groovy = 'SLASHSTAR_STYLE'
     }
-    ext.year = '2017-2020'
+    ext.year = '2017-2022'
 
     exclude "**/transaction/**"
     exclude '**/*.txt'
--- a/openapi/openapi-custom-schema-for-class.properties
+++ b/openapi/openapi-custom-schema-for-class.properties
@@ -1,0 +1,4 @@
+micronaut.openapi.schema.io.micronaut.openapi.ObjectId=java.lang.String
+micronaut.openapi.schema.io.micronaut.openapi.JAXBElement=io.micronaut.openapi.MyJaxbElement
+micronaut.openapi.schema.io.micronaut.openapi.JAXBElement<test.XmlElement2>=io.micronaut.openapi.MyJaxbElement2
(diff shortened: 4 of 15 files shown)


altro3:
@graemerocher Hi! could you merge and release this very important feature?

graemerocher:
these should be specified as annotation processor inputs otherwise the compiler will not pick up the changes if you change any of the values.

altro3:
@graemerocher Sorry, I didn't understand what you mean. My idea is that you specify in the configuration which classes in the API structure you want to replace with other classes. The compiler is not required for this action, all class structures are loaded from the context. In addition, I tested all these changes locally on my project. I still don't understand what could be the problem.

graemerocher:
in general this module does have problems with incremental compile since everything seems to be configured via `openapi.properties` so maybe this is a broader concern. The issue though is that if you change `openapi.properties` and the sources haven't changed, running the compilation (without first doing a clean) will not recompile the classes in order to pick up the changes, unless you supply the correct inputs to the compiler through the use of annotation processor arguments.

altro3:
Hmm... Could you show an example how to check if you are right or not. And maybe you can suggest how to fix this problem.

In general, this feature turned out to be really very important. In my project, I ran into the problem of generating JAXBElement<? extends SomeGeneratedClass>. I added a custom serializer to jackson so that the field data is correctly represented in json, but I realized that I can’t generate the correct swagger file in any way, because. The code is generated and I can't change it. With the help of this feature, it will be possible to fix this problem

altro3:
@graemerocher Hi! So what do you end up suggesting? This feature is very important and without it at the moment it is impossible to work with the generated files. If there are suggestions on how to improve my solution - no problem, I will do it.

graemerocher:
Looks like the feature is missing documentation in general, so please add documentation and in the documentation state that changing the configuration requires a full recompile. Thanks.

altro3:
@graemerocher done

altro3:
@graemerocher Hi! Sorry to rush, but this feature is really very important. In addition, I found a small bug and it can also be included in the release. Could you merge everything and release it?

graemerocher:
since this is a new feature could you bump the minor version and update the target branch to 3.8.x

Thanks

altro3:
@graemerocher done!

Issue #847, opened 2022-11-16
@Schema properties are not reflected in swagger api


kevind-wgu:
### Feature description

From a previous conversation I was trying to figure out how to get a Value class to show up as a String. I was recommend to use `@Schema(implementation=String.class)` as a way to manually define a primitive type for an object.

This works great, however the properties that you can specify on `@Schema` are not transferred to the the swagger.yml properties.

Example:
```
import com.fasterxml.jackson.annotation.JsonCreator;
import com.fasterxml.jackson.annotation.JsonValue;
import io.micronaut.core.annotation.Introspected;
import io.swagger.v3.oas.annotations.media.Schema;
import lombok.Value;

import javax.validation.constraints.NotNull;
import javax.validation.constraints.Pattern;

@Value
@Introspected
@Schema(implementation = String.class, pattern = EventType.VALUE_PATTERN, minLength = 4, maxLength = 40)
public final class EventType {
    static final String VALUE_PATTERN = "[a-zA-Z1-9._\\-@]{4,40}";

    @NotNull
    @Pattern(regexp = VALUE_PATTERN)
    private final String value;

    public EventType(String value) {
        this.value = value;
    }

    @JsonCreator
    public static EventType of(String value) {
        return new EventType(value);
    }

    @JsonValue
    public String getValue() {
        return value;
    }

    @Override
    public String toString() {
        return value;
    }
}

import io.micronaut.core.annotation.Introspected;
import javax.validation.constraints.NotNull;
import javax.validation.constraints.Pattern;
import lombok.Data;

@Data
@Introspected
public class EventTypeHolder {
    private EventType type1;
    @Pattern(regexp = "[a-zA-Z1-9._\\-@]{4,40}")
    private String type2;
}

```

This results in the following JSON schema.  Take note how the "type2" property has the pattern defined, but the "type1" property does not even though I manually defined it on the `@Schema`.

```
components:
  schemas:
    EventTypeHolder:
      required:
      - type2
      type: object
      properties:
        type1:
          type: string
        type2:
          pattern: "[a-zA-Z1-9._\\\\-@]{4,40}"
          type: string
```

This only appears to be a problem on primitive types. 

I would appreciate it if when overwriting the "implementation" of a given class for swagger via `@Schema` to a primitive type that the other properties on `@Schema` be used on the resulting schema.

-----------------

Debugging the code it appears that for me the schema is generated in `AbstractOpenApiVisitor.resolveSchema:819`.
```
[...]

PR #849, opened 2022-11-16
Apply Schema Attributes to subtitute Schema

kevind-wgu:
This is attempting to address issue #847. If you would like me to adjust things please let me know.

When specifying @Schema(implementation=String.class ...) all of the other parameters on the annotation are lost. The changes here will apply many of those annotation parameter values to the substitute Schema. The ones I applied are pattern, maximum, minimum, maxLength, minLength, exclusiveMaximum, exclusiveMinimum, example, defaultValue, multipleOf.


altro3:
@kevind-wgu hi! thanks for help! I fix this bug here: #851. My changes base on your PR. But we already have `@Schema` annotation processing. You don't need to dubplicate logic for this

kevind-wgu:
Sounds great! Thank you!

PR #851, opened 2022-11-18
Added processing Schema attributes, when it used with implementation.

### LINKED ISSUE #847: @Schema properties are not reflected in swagger api

altro3:
Based on #849.

Fixed #847

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ projectVersion=4.7.1-SNAPSHOT
 projectGroup=io.micronaut.openapi
 
 micronautDocsVersion=2.0.0
-micronautVersion=3.7.3
+micronautVersion=3.7.4
 micronautTestVersion=3.7.0
 groovyVersion=3.0.13
 spockVersion=2.3-groovy-3.0
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
-managed-swagger = "2.2.6"
+managed-swagger = "2.2.7"
 # Required to keep catalog compatibility with 3.4.x.  Can be removed for 4.0.0
-managed-swagger-compat = "2.2.6"
+managed-swagger-compat = "2.2.7"
 managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.15.3"
 managed-jakarta-validation-api = "3.0.2"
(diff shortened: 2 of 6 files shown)


altro3:
@graemerocher Hi! Could you merge and release this PR and #845 ? I'm sorry that this happened, but could you create a release today or tomorrow?

graemerocher:
Thanks for the contribution!

PR #853, opened 2022-11-20
Add correct processing exploded QueryValues

### LINKED ISSUE #852: Micronaut OpenAPI spec isn't generated correctly
### Expected Behavior

It seems the generation of OpenAPI specifications doesn't take [RFC-6570 URI template specification](https://tools.ietf.org/html/rfc6570) into account.

I have a method in Micronaut which looks like this:

`fun checkAvailability(availabilityRequest: AvailabilityRequest): Availability {`

My request looks like this:

```
data class AvailabilityRequest(
    val restaurantId: String,
    val partySize: Int,
    val visitTime: Instant,
)
```

However, I don't want to pass in a request, I'd like to pass in the individual parameters, such as:

`http://localhost/availability?restaurantId=1&partySize=2&visittime=whenever`

To do this, I annotated the controller method with this annotation:

`@Get("/{?availabilityRequest*}")`

(based on the info I found here: https://dzone.com/articles/micronaut-mastery-binding-request-parameters-to-po)

This works well, and all good. But now I want to generate an OpenAPI specification based on the above, and the generation doesn't seem to take the `?availabilityRequest*` into account. Instead, it still thinks the availabilityRequest is a parameter, which makes my mocks fail. 

The generated OpenAPI spec looks like this:

```
paths:
  /availability:
    get:
      tags:
      - availability
      summary: Availability check
      description: Returns the available timeslots
      operationId: checkAvailability
      parameters:
      - name: availabilityRequest
        in: query
        required: true
        explode: true
        schema:
          $ref: '#/components/schemas/AvailabilityRequest'
```

While I expected the OpenAPI spec to look more like this:

```
paths:
  /availability:
    get:
      tags:
      - availability
      summary: Availability check
      description: Returns the available timeslots
      operationId: checkAvailability
      parameters:
      - name: restaurantId
        in: query
        required: true
        schema:
          type: string
      - name: partySize
        in: query
        required: true
        schema:
          type: integer
          format: int32
```

### Actual Behaviour

See above

### Steps To Reproduce

Create a new app with a complex type as input for a Get method (maybe also Post?)
Enable openapi
Build the spec

### Environment Information

MacOS

### Example Application

_No response_

### Version

3.7.4

altro3:
Fixed #852

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
@@ -640,29 +640,53 @@ public abstract class AbstractOpenApiEndpointVisitor extends AbstractOpenApiVisi
         if (newParameter == null) {
             return;
         }
-        if (StringUtils.isEmpty(newParameter.getName())) {
-            newParameter.setName(parameter.getName());
-        }
 
-        if (newParameter.getRequired() == null && !isNullable(parameter)) {
-            newParameter.setRequired(true);
-        }
-        if (javadocDescription != null && StringUtils.isEmpty(newParameter.getDescription())) {
-            CharSequence desc = javadocDescription.getParameters().get(parameter.getName());
-            if (desc != null) {
-                newParameter.setDescription(desc.toString());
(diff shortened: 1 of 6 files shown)


graemerocher:
Thanks for the contribution!

bodiam:
Oh wow, that's an incredibly quick fix! Thank you!

PR #884, opened 2022-12-12
Fixed bug with wrong references for schemas with inheritance, when used custom schema name

altro3:
Fixed bug with wrong references for schemas with inheritance, when used custom schema name


altro3:
@graemerocher Hi! Could you merge ans release these PRs: #882, #884, #881 ?

This fix very important.

PR #885, opened 2022-12-13
New feature: schema decorators

altro3:
If you have some classes with same names in different packages you can set postfix like this:

```yaml
micronaut:
  openapi:
    schema-postfix:
      org.api.v1_0_0: 1_0_0
      org.api.v2_0_0: 2_0_0
```
or by system properties:
```
-Dmicronaut.openapi.schema-postfix.org.api.v1_0_0=1_0_0 -Dmicronaut.openapi.schema-postfix.org.api.v2_0_0=2_0_0
```
or by openapi.properties
```properties
micronaut.openapi.schema-postfix.org.api.v1_0_0=1_0_0
micronaut.openapi.schema-postfix.org.api.v2_0_0=2_0_0
```

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -2121,9 +2121,11 @@ abstract class AbstractOpenApiVisitor {
         if (metaAnnName != null && !io.swagger.v3.oas.annotations.media.Schema.class.getName().equals(metaAnnName)) {
             return NameUtils.getSimpleName(metaAnnName);
         }
+        String fullClassName;
         String javaName;
         if (type instanceof TypedElement) {
             ClassElement typeType = ((TypedElement) type).getType();
+            fullClassName = typeType.getName();
             if (CollectionUtils.isNotEmpty(typeType.getTypeArguments())) {
                 javaName = computeNameWithGenerics(typeType, typeArgs, context);
             } else {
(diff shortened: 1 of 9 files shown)


altro3:
@graemerocher Hi! Could you merge this PR and release a new 4.9.0 release?

I ran into a problem on my project. I have generated API models according to the xsd scheme. There can be up to 10 such schemes, each for its own version of the API. The problem is that micronaut-openapi does not take into account when generating schemas that the classes are in different packages, but the classes are named the same. Because of this, the final swagger file is generated incorrectly.

This is one of the methods for solving this problem, suitable for me. In the future, I will add the ability to specify a behavior strategy for classes with the same name, i.e. so that, for example, an index is automatically added to the end of the model name, as is now done with duplicate operationId

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -2121,18 +2121,30 @@ private String computeDefaultSchemaName(Element definingElement, Element type, M
...
                 javaName = computeNameWithGenerics(typeType, Collections.emptyMap(), context);
             }
         } else {
             javaName = type.getSimpleName();
+            fullClassName = type.getName();
         }
-        return javaName.replace("$", ".");
+        String packageName = fullClassName.substring(0, fullClassName.lastIndexOf('.'));
+        OpenApiApplicationVisitor.SchemaDecorator schemaDecorator = OpenApiApplicationVisitor.getSchemaDecoration(packageName, context);
+        javaName = javaName.replace("$", ".");
```
graemerocher, line 2140: you should use `getCanonicalName()` on the element instead of this hack

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
```diff
@@ -218,6 +218,25 @@ public class OpenApiApplicationVisitor extends AbstractOpenApiVisitor implements
      */
     private static final String MICRONAUT_OPENAPI_SCHEMA = "micronaut.openapi.schema";
     private static final String MICRONAUT_CUSTOM_SCHEMAS = "micronaut.internal.custom.schemas";
+    /**
+     * Properties prefix to set schema name prefixe or postfix by package.
```
graemerocher, line 222: typo `prefixe`

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
altro3: What do you mean? If you are talking about substring by dollar symbol, then this is not my code, it was written that way. And I didn't understand how you suggest to use canoncialName ? getCanoncialName returns the full name of the class along with the package, but I don't need it, I need either the class name or the package name.

In any case, I changed the logic for calculating the package name.

altro3:
@graemerocher Thanks! :-) Could you release all fixes ?

graemerocher:
Done, thanks for the contributions!

Issue #878, opened 2022-12-06
Swagger UI resources cannot be loaded behind a reverse proxy (with path prefix strip)


srettenmeier:
### Expected Behavior

When running an app behind a reverse proxy, that is also stripping path prefixes, 
the Swagger UI resources (JS, CSS) should be loaded taking the stripped prefix into account.

### Actual Behaviour

With version 4.7.1, the generated Swagger UI's `index.html` references resources absolutely (e.g. `<script src='/swagger-ui/res/swagger-ui-bundle.js'>`). With these references in place the browser cannot load the resources, as the stripped path prefix is not taken into account.

Prior, we used version 4.5.2. There the references where relative (e.g. `<script src="res/swagger-ui-bundle.js">`). This way the resources could also be loaded behind a reverse proxy

### Steps To Reproduce

1. Start an arbitrary Micronaut app with a Swagger UI (generated by Gradle kapt processor)
2. Set up a reverse proxy (e.g. nginx)
3. Configure the reverse proxy to strip a certain prefix (e.g. `/sample-app`) and forward requests to the Micronaut app
4. Open the Swagger UI in your browser by using the reverse proxy's URL including the path prefix (e.g. `http://<reverse proxy address>/sample-app`)

### Version

4.7.1

altro3:
Hi! This is not an error, but you can change the path to the resources as you need using the property `swagger-ui.js.url`. 
```groovy
options.forkOptions.jvmArgs += "-Dmicronaut.openapi.views.spec=swagger-ui.enabled=true,swagger-ui.js.url=res/"
```

altro3:
But I found a different bug with *.js.url properties. Fix is here #879

srettenmeier:
Hi @altro3,

first of all, thanks for the super fast feedback! 


> Hi! This is not an error, but you can change the path to the resources as you need using the property `swagger-ui.js.url`.
> 
> ```groovy
> options.forkOptions.jvmArgs += "-Dmicronaut.openapi.views.spec=swagger-ui.enabled=true,swagger-ui.js.url=res/"
> ```

That's what I tried, but it didn't have the desired effect. The resources were still referenced absolutely (e.g. <script src='/swagger-ui/res/swagger-ui-bundle.js'>).

> But I found a different bug with *.js.url properties. Fix is here #879

I'm glad you found out that this is a bug and already fixed it.

Thank you so much!

altro3:
[...]

PR #896, opened 2023-01-04
Fixed setting server context with url parameter.

ondrej-hrstka:
Hi, this PR fixes broken setting of the [context path by the url param](https://micronaut-projects.github.io/micronaut-openapi/latest/guide/#urlParameter). Please note that, it changes the default path to resources to relative one.


altro3:
@ondrej-hrstka Hi! This is right default behavior. If you need to change default path, just use property:
```groovy
options.forkOptions.jvmArgs += "-Dmicronaut.openapi.views.spec=swagger-ui.enabled=true,swagger-ui.js.url=res/"
```
See this: https://github.com/micronaut-projects/micronaut-openapi/issues/878

PR #899, opened 2023-01-07
Actualize libraries versions. Added accepted-api-changes lines. Swagger 2.2.8

altro3:
Actualize libraries versions. Added accepted-api-changes lines. Swagger 2.2.8

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,21 +1,19 @@
 [versions]
-managed-swagger = "2.2.7"
-# Required to keep catalog compatibility with 3.4.x.  Can be removed for 4.0.0
-managed-swagger-compat = "2.2.7"
+managed-swagger = "2.2.8"
 managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.15.3"
 managed-slf4j = "2.0.6"
 managed-html2md-converter = "0.64.0"
 managed-jakarta-validation-api = "3.0.2"
 
-kotlin = "1.7.22"
+kotlin = "1.8.0"
 micronaut-security = "4.0.0-SNAPSHOT"
 micronaut-serde = "2.0.0-SNAPSHOT"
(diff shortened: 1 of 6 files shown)


altro3:
@graemerocher Need help.  Don't understand how I can delete old library version `swagger-compat`:
```
 Execution failed for task ':openapi-bom:checkVersionCatalogCompatibility'.
> Version catalogs are not compatible:
  The following versions were present in the baseline version but missing from this catalog:
    - swagger-compat
  The following libraries were present in the baseline version but missing from this catalog:
    - swagger
```

graemerocher:
@melix can you help explain how a library can be excluded from this check

gradle/libs.versions.toml
```diff
@@ -1,21 +1,19 @@
 [versions]
-managed-swagger = "2.2.7"
-# Required to keep catalog compatibility with 3.4.x.  Can be removed for 4.0.0
-managed-swagger-compat = "2.2.7"
+managed-swagger = "2.2.8"
```
melix, line 2: :x: It seems to be a mistake to keep `managed-swagger` in the versions, when the _library_ was removed below.

melix (changes requested):
You can accept the catalog breaking changes by adding this in the `bom` build file:

```
micronautBom {
    suppressions {
        acceptedVersionRegressions.addAll(["managed-swagger", "managed-swagger-compat"])
        acceptedLibraryRegressions.add("managed-swagger")
}
```

gradle/libs.versions.toml
altro3: @melix thanks for help! Now all works fine. Right solution is:
```groovy
micronautBom {
    suppressions {
        acceptedVersionRegressions.add("swagger-compat")
        acceptedLibraryRegressions.add("swagger")
    }
}
```
Need to remove library `swagger`, because right name now is `swagger-core`

melix (approved):

PR #909, opened 2023-01-13
Fix compilation issue with MultipartBody

### LINKED ISSUE #907: Compilation issue with MultipartBody
### Expected Behavior

Compilation should end well, without issue, when a controller uses a MultipartBody body

### Actual Behaviour

When using a `MultipartBody` body (with `@Body`) in a controller, compilation fails in `AbstractOpenApiVistor#bindSchemaForElement` because schema is null.  
And Schema is null because the schema resolution for the body checks if the body is a container type and use its first type argument instead.
`MultipartBody` is considered as a container type as it extends `Publisher<CompletedPart>` but `getFirstTypeArgument` returns nulls because itself doesn't have any argument.

A workaround is to use `Publisher<CompletedPart>` instead of `MultipartBody`.

### Steps To Reproduce

See example

### Environment Information

_No response_

### Example Application

https://github.com/ggerbaud/demo-openapi-multipartbody

### Version

micronaut-openapi 4.8.1

ggerbaud:
fix #907

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -779,9 +779,18 @@ abstract class AbstractOpenApiVisitor {
             boolean isPublisher = false;
             boolean isObservable = false;
             boolean isNullable = false;
-
+            // MultipartBody implements Publisher<CompletedPart> (Issue : #907)
+            if (type.isAssignable("io.micronaut.http.server.multipart.MultipartBody")) {
+                isPublisher = true;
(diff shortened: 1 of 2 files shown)


graemerocher:
@altro3 could you review this one. Also we should probably discuss whether you would like commit rights on this repository since you are contributing so much. It will make it easier for you to assign issues, close issues etc.

altro3:
@graemerocher Hi! Yes, this PR is good and need to merge it to 4.8.2. I analyzed out all the issues that have been created lately. After merging this PR, you can safely release 4.8.2.

About granting rights: yes, I think it would be easier for everyone :-)

altro3:
@graemerocher Could you merge this PR and finalize release 4.8.2?

Issue #912, opened 2023-01-17
Incorrect warning during openapi generation: 'name' property can't set for securityScheme with type http.


StefanSrf:
### Expected Behavior

no warning is logged during the build

### Actual Behaviour

the following warning is logged when generating the openAPI documentation:
warning: 'name' property can't set for securityScheme with type http. Skip it

Note: the generated openapi yml file is correct, the security scheme with the given name is added to the yml. If the 'name' value is removed from the securityScheme the warning disappears, but then the security scheme is no longer added to the yml.

### Steps To Reproduce

Add the following to the Application

@SecurityScheme(
    name = "basicAuth",
    scheme = "Basic",
    type = SecuritySchemeType.HTTP,
)

build the project

### Environment Information

Ubuntu 22.04
JDK 17

### Version

3.8.0

altro3:
@StefanSrf already fixed here: https://github.com/micronaut-projects/micronaut-openapi/pull/892

Wait release 4.8.2

Issue #911, opened 2023-01-14
kotlinKapt crashes when a component or value is injected into the controller


darylrobbins:
### Expected Behavior

Project should build successfully

### Actual Behaviour

After calling `gradlew clean build` I receive the following error:

```
> Task :kaptKotlin FAILED
5 actionable tasks: 5 executed
Note: Generating OpenAPI Documentation
error: Error finalizing type visitor [io.micronaut.openapi.visitor.OpenApiApplicationVisitor@7787e4cf]: null
Note: Creating bean classes for 3 type elements
error: Error finalizing type visitor [io.micronaut.openapi.visitor.OpenApiApplicationVisitor@7787e4cf]: null

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kaptKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask$KaptExecutionWorkAction
   > java.lang.reflect.InvocationTargetException (no error message)
```

### Steps To Reproduce

I started from the test case for #903 and modified to isolate this issue.

Application Type: Micronaut Application
Java version: 17
Micronaut Version: 3.8.0
Language: Kotlin
Build Tool: Gradle Kotlin
Test Framework: Kotest
In features: swagger-ui and openapi

I define a controller with either an injected component or value.

```
@Controller("/demo")
class DemoController(
    // Fails with either of these two present:
    @Inject val demoService: DemoService, // (1)
    @Value("\${baseurl:}") val baseUrl: String, // (2)
) { ... }
```

It will fail with either (1) or (2) commented out.

Full gradle stacktrace in attachment.
[stacktrace.txt](https://github.com/micronaut-projects/micronaut-openapi/files/10417999/stacktrace.txt)

### Environment Information

- Operating System: Mac OS
- Java: Amazon Corretto 17 (but happens on OpenJDK as well)

### Example Application

https://github.com/darylrobbins/micronaut-openapi-test

### Version

3.8.0

altro3:
@darylrobbins Hi! Already fixed. Checked with version 4.8.2-SNAPSHOT. This is the same problem as here https://github.com/micronaut-projects/micronaut-openapi/issues/903 - you return ArrayList, not List

Issue #903, opened 2023-01-12
kotlinKapt crash in some cases when use ArrayList<T> or ArrayDeque<T> as return type in controller


marendil:
### Expected Behavior

Project should build successfully.

### Actual Behaviour

After calling `gradlew clean build` I've got following error:
```
> Task :kaptKotlin
Note: Generating OpenAPI Documentation
error: Error finalizing type visitor [io.micronaut.openapi.visitor.OpenApiApplicationVisitor@107268a5]: null
Note: Creating bean classes for 2 type elements
error: Error finalizing type visitor [io.micronaut.openapi.visitor.OpenApiApplicationVisitor@107268a5]: null

> Task :kaptKotlin FAILED

FAILURE: Build failed with an exception.
```

### Steps To Reproduce

I've created new project in Micronaut Launcher https://micronaut.io/launch/ with following options:

Application Type: Micronaut Application
Java version: 17
Micronaut Version: 3.8.0
Language: Kotlin
Build Tool: Gradle Kotlin
Test Framework: Kotest
In features: swagger-ui and openapi

Next I've created new simple data class:

`data class AC(val someField: String)`

and changed index method so it returns empty `ArrayList<AB>`

`fun index() = ArrayList<AB>()`

Then I call  `gradlew clean build` and it fails.

What is funny following class names works:

```
A
Ab
SomeClass
Xbcd
```

but following class names are not:
```
Abc
AB
ABc
AwesomeClass
```

what is more consusing is that it always works for List or MutableList but doesn't work for ArrayList and ArrayDeque

It starts to crash from version 4.8.0 of micronaut-openapi, after downgrading to 4.7.1 it works like a charm.

Full gradle stacktrace in attachment.
[fullstacktrace.txt](https://github.com/micronaut-projects/micronaut-openapi/files/10401018/fullstacktrace.txt)

### Environment Information

- JDK 17
- Gradle wrapper 7.5.1

### Example Application

https://github.com/marendil/demo

### Version

3.8.0

altro3:
Fixed here: https://github.com/micronaut-projects/micronaut-openapi/pull/905

Colonel-Rogue:
This also occurs when returning a [kotlin Flow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/) from a controller function.
[...]

PR #915, opened 2023-01-24
Add getting schema type from JsonValue method and fixed incorrect processing supertypes for enums

altro3:
Don't need to create schemas and process them for classes `java.lang.Object` and `java.lang.Enum`

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -69,11 +69,13 @@ import io.micronaut.http.MediaType;
 import io.micronaut.http.uri.UriMatchTemplate;
 import io.micronaut.inject.ast.ClassElement;
 import io.micronaut.inject.ast.Element;
+import io.micronaut.inject.ast.ElementQuery;
 import io.micronaut.inject.ast.EnumConstantElement;
 import io.micronaut.inject.ast.EnumElement;
 import io.micronaut.inject.ast.FieldElement;
 import io.micronaut.inject.ast.GenericPlaceholderElement;
 import io.micronaut.inject.ast.MemberElement;
+import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.ast.PropertyElement;
 import io.micronaut.inject.ast.TypedElement;
 import io.micronaut.inject.visitor.VisitorContext;
@@ -116,6 +118,7 @@ import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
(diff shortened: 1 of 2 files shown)


altro3:
@graemerocher Hi! Could you merge this PR ? Sorry, found this bug only after 4.8.2 has released :-(. 

Also could you close these isssues:

https://github.com/micronaut-projects/micronaut-openapi/issues/900
https://github.com/micronaut-projects/micronaut-openapi/issues/903
https://github.com/micronaut-projects/micronaut-openapi/issues/907
https://github.com/micronaut-projects/micronaut-openapi/issues/911
https://github.com/micronaut-projects/micronaut-openapi/issues/912

They all fixed in 4.8.2.

openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiEnumSpec.groovy
```diff
@@ -1,12 +1,134 @@
...
+    SECOND(2),
+    ;
+
+    private final int value;
+
+    MyEnum(int value) {
+        this.value = value;
+    }
+
+    @JsonValue
```
sdelamo, line 44: This test is confusing to me. What are we testing @JsonProperty or @JsonValue?

should not we have two tests?

```
enum MyEnum {
     FIRST(1),
     SECOND(2) ;

     private final int value;

     MyEnum(int value) {
         this.value = value;
     }

     @JsonValue
     public int getValue() {
         return value;
     }
}
```` 

```
enum MyEnum {
    @JsonProperty("1")
     FIRST(1),
    @JsonProperty("2")
     SECOND(2) ;
}
````

**Review by sdelamo — CHANGES_REQUESTED** [review]

File: openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiEnumSpec.groovy (hunk 1 above)
Comment (altro3) [reply openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiEnumSpec.groovy:44]: @sdelamo No, you missunderstood. We need `@JsonProperty` annotation to set enum values which different with enum value name. This is single way to customize enum values. In tests we need to set int values for every enum value and annotation processor haven't  acccess to enum valus objects, only to structure and that's why we need to add annotation JsonProperty

Comment (sdelamo) [reply openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiEnumSpec.groovy:44]: > We need @JsonProperty annotation to set enum values which different with enum value name.

But you are using the exact value not a different one. 

The test has: 

```
    @JsonProperty("1")
     FIRST(1),
     @JsonProperty("2")
     SECOND(2),
```

It does not have:


```
    @JsonProperty("5")
     FIRST(1),
     @JsonProperty("6")
     SECOND(2),
```


Moreover, are my examples not how people will write code in their applications? You will never write an enum with both `@JsonProperty` and `@JsonValue`  right?

Comment (altro3) [reply openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiEnumSpec.groovy:44]: @sdelamo not exact. Value for enum value `FIRST` will be "FIRST", not 1. Again: annotation process can't see parameters in construct that you set. In other words, micronaut openapi don't see constructor argument value 1 or 2, micronaut-openapi can see only annotaions parameters

File: openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java [hunk 2]
```diff
@@ -2072,6 +2088,24 @@ protected Schema readSchema(AnnotationValue<io.swagger.v3.oas.annotations.media.
         return schema;
     }
 
+    private String checkEnumJsonValueType(EnumElement type, String schemaType) {
```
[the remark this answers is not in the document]
altro3: done

openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiEnumSpec.groovy
altro3: @sdelamo 
```java
@JsonProperty("1")
FIRST(1),
@JsonProperty("2")
SECOND(2)
```
This need to set number values for FIRST and SECOND enum values. Because this is sungle way to set numbers to enum values in annotation processor. The other way is much more complicated and nned to run the compiler directly from the annotation processor

altro3: And yes, this is a limitation. You will have to declare both JsonProperty and JsonValue if you want to use micronaut-openapi . These are the limitations of the annotation processor. Another way is to explicitly specify the type in the @Schema annotation.

For example I'm using JsonValue and JsonProperty in the generated code by hud schema via xjc .

PR #917, opened 2023-01-24
Correct how non-default annotation values are retrieved

dstepanov:
After changes https://github.com/micronaut-projects/micronaut-core/pull/8632 that are coming in Micronaut 4.
Default annotations values are preserved for source annotations. This PR corrects how the API should be used to avoid the default values.


dstepanov:
FYI @altro3

graemerocher (approved):

PR #923, opened 2023-01-31
Fixed process nullable annotations

### LINKED ISSUE #921: Jakarta's annotations are ignored
### Expected Behavior

When using annotations from the `jakarta.validation.*` and `jakarta.annotations.*` package they should behave the same as from `javax.*` and produce oas that includes the annotations.

### Actual Behaviour

The annotations are ignored.

### Steps To Reproduce

1. Create a project with an API endpoint that takes an Pojo as parameter
2. use `@Valid` on that parameter
3. create field in the   `@jakarta.annotations.Nullable @jakarta.validation.Size(min=1,max=2) public String a;`
4. inspect the generated yml file  

### Environment Information

Any environment.

Maybe this should be done in sync with: https://github.com/micronaut-projects/micronaut-core/issues/4770 but it might also be a good idea to support it before micronaut core does it. At least for `@Nullable` since Nullable doesn't create any behavior but just adds intellisense.

I would argue that every annotation `*.Nullable` should be considered since there are 100 and one (Non)Nullability annotation providers like `jetbrains`, `findbugs`, `spotbugs` and so on

### Example Application

_No response_

### Version

`3.*.*`

altro3:
Fixed #921

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,6 +6,10 @@ managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.15.3"
 managed-jakarta-validation-api = "3.0.2"
 managed-slf4j = "1.7.36"
+managed-jspecify = "0.3.0"
+managed-jdt-annotation = "2.2.700"
+managed-android-annotation = "1.5.0"
+managed-spotbugs-annotations = "4.7.3"
 
 # Versions beyond 0.62.2 require Java 11
 managed-html2md-converter = "0.62.2"
@@ -36,6 +40,10 @@ jackson-datatype-jsr310 = { module = "com.fasterxml.jackson.datatype:jackson-dat
 jackson-dataformat-yaml = { module = "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml" }
 javax-validation-api = { module = "javax.validation:validation-api" }
 jakarta-validation-api = { module = "jakarta.validation:jakarta.validation-api", version.ref = "managed-jakarta-validation-api" }
+spotbugs-annotations = { module = "com.github.spotbugs:spotbugs-annotations", version.ref = "managed-spotbugs-annotations" }
+jspecify = { module = "org.jspecify:jspecify", version.ref = "managed-jspecify" }
+jdt-annotation = { module = "org.eclipse.jdt:org.eclipse.jdt.annotation", version.ref = "managed-jdt-annotation" }
+android-annotation = { module = "androidx.annotation:annotation", version.ref = "managed-android-annotation" }
 
 testcontainers-spock = { module = "org.testcontainers:spock" }
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,17 +4,17 @@ managed-swagger = "2.2.7"
 managed-swagger-compat = "2.2.7"
 managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.15.3"
-managed-jakarta-validation-api = "3.0.2"
 managed-slf4j = "1.7.36"
-managed-jspecify = "0.3.0"
-managed-jdt-annotation = "2.2.700"
-managed-android-annotation = "1.5.0"
-managed-spotbugs-annotations = "4.7.3"
 
 # Versions beyond 0.62.2 require Java 11
 managed-html2md-converter = "0.62.2"
 
 kotlin = "1.7.22"
+jakarta-validation-api = "3.0.2"
+jspecify = "0.3.0"
+jdt-annotation = "2.2.700"
+android-annotation = "1.5.0"
+spotbugs-annotations = "4.7.3"
 
 [libraries]
 # Duplicated to keep catalog compatibility with 3.4.x.  Can be removed for 4.0.0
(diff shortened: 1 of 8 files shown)


nbrugger-tgm (approved):
Loogs good to me, maybe pattern matching (like `*.Nullable`) could be considered but that could have unintended side effects in very rare cases. But that's up to you / just an idea

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -2462,4 +2460,12 @@ protected <T, A extends Annotation> List<T> processOpenApiAnnotation(Element ele
         }
         return tagList;
     }
+
+    public boolean isNullable(TypedElement element) {
```
sdelamo, line 2464: Please move this method to a `TypeElementUtils` class.

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -2462,4 +2460,12 @@ protected <T, A extends Annotation> List<T> processOpenApiAnnotation(Element ele
...
+        return element.isNullable()
+            || element.getType().isOptional()
+            || element.hasStereotype(Nullable.class)
+            || element.hasStereotype("jakarta.annotation.Nullable")
+            || element.hasStereotype("org.jetbrains.annotations.Nullable");
```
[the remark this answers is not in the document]
dstepanov: This can be implemented using an annotation transformer to transform the annotations to `AnnotationUtil.NULLABLE`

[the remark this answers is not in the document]
altro3: add all other annotations too

openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiNullableTypesSpec.groovy
```diff
@@ -257,4 +257,72 @@ class PetController {
...
+
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.http.annotation.Body;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Get;
+
+@Introspected
+class HelloWorldDto {
+    @jakarta.annotation.Nullable
```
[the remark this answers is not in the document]
altro3: I did not understand how to change the test. I'm sorry, I don't have much experience with Spock.

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
altro3: done

altro3:
@sdelamo @dstepanov I've added support all nullable annotations from @sdelamo 's link. But I think i found bug in micronaut: 
```
element.hasStereotype("org.eclipse.jdt.annotation.Nullable")
element.hasStereotype("org.jspecify.annotations.Nullable")
```
These two annotations not found.

I think problem is here: both annotions have Target TYPE_USE
```
@Target(TYPE_USE)
```

openapi/src/main/java/io/micronaut/openapi/visitor/TypeElementUtils.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param element typed element
+     *
+     * @return true if element is nullable, false - otherwise.
+     */
+    public static boolean isNullable(TypedElement element) {
+        return element.isNullable()
+            || element.getType().isOptional()
+            || element.hasStereotype(Nullable.class)
+            || element.hasStereotype("javax.annotation.Nullable")
```
dstepanov, line 27: Can you use an annotation transformer https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/core-processor/src/main/java/io/micronaut/inject/annotation/internal/JakartaNullableTransformer.java ?

dstepanov (changes requested):

dstepanov:
@altro3 Javac might be acting strange with those types. Maybe you can create an issue with a reproducible test in the core?

openapi/src/main/java/io/micronaut/openapi/visitor/TypeElementUtils.java
altro3: @dstepanov No, can't. Because it since micronaut 4.0, but this fix for release 4.8.3 which based on micronaut 3.8.3

dstepanov: It should work in 3.8.3 too.

openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiNullableTypesSpec.groovy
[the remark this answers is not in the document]
sdelamo: I have done a commit with a more idiomatic way of testing this with Spock 

https://github.com/micronaut-projects/micronaut-openapi/pull/923/commits/0ceee653914162b9a3b2c517968815db9265e7c1

gradle/libs.versions.toml
```diff
@@ -6,6 +6,10 @@ managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.15.3"
 managed-jakarta-validation-api = "3.0.2"
 managed-slf4j = "1.7.36"
+managed-jspecify = "0.3.0"
```
sdelamo, line 9: I don't think we should manage these dependencies. We should not publish them in the openAPI module's BOM.

sdelamo (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/TypeElementUtils.java
```diff
@@ -0,0 +1,38 @@
+package io.micronaut.openapi.visitor;
```
sdelamo, line 1: it misses the license

sdelamo (changes requested):

altro3:
@dstepanov Done. Look here:

https://github.com/micronaut-projects/micronaut-core/pull/8707
https://github.com/micronaut-projects/micronaut-core/issues/8706

gradle/libs.versions.toml
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/TypeElementUtils.java
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/TypeElementUtils.java
altro3: @dstepanov no! this class added in micronaut 4.0:
изображение

Can;t find this class in 3.8.x branch

gradle/libs.versions.toml
```diff
@@ -4,13 +4,17 @@ managed-swagger = "2.2.7"
 managed-swagger-compat = "2.2.7"
 managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.15.3"
-managed-jakarta-validation-api = "3.0.2"
```
sdelamo, line 7: if we were already managing jakarta we should not remove it. That could be a breaking change for users' applications.

sdelamo (changes requested):

gradle/libs.versions.toml
altro3: @sdelamo ok. I reverted it. But it was mistake. We don't need to manage this dependency too

gradle/libs.versions.toml
```diff
@@ -11,6 +11,11 @@ managed-slf4j = "1.7.36"
 managed-html2md-converter = "0.62.2"
 
 kotlin = "1.7.22"
+jakarta-validation-api = "3.0.2"
```
sdelamo, line 14: remove unused

sdelamo (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/TypeElementUtils.java
dstepanov: Yes, the class is just an example how to do it. Just copy it, modify the annotation name and define it in the service entires.

sdelamo (approved):

PR #925, opened 2023-02-02
ci: Github Actions sync

sdelamo:


.github/workflows/graalvm.yml
```diff
@@ -42,7 +42,11 @@ jobs:
           version: ${{ matrix.graalvm }}
           java-version: ${{ matrix.java }}
           components: 'native-image'
```
timyates, line 44: This only runs Graal 17 is that right?

timyates (approved):

.github/workflows/graalvm.yml
sdelamo: Yes

```

    strategy:
      matrix:
        java: ['17']
        graalvm: ['latest']
```

PR #930, opened 2023-02-10
Fixed wrong byte type detection. Again remove generation schema for class java.lang.Object

altro3:


altro3:
@graemerocher Hi! Could you merge and release this fix? The problem currently occurs when enums are converted to type byte . An example of such an enum:
```java
enum Type1 {
    @JsonProperty("1")
    _1(((byte) 1)),
    @JsonProperty("2")
    _2(((byte) 2));
    private final byte value;
    Type1(byte value) {
        this.value = value;
    }
    @JsonValue
    public byte value() {
        return value;
    }
    @JsonCreator
    public static Type1 fromValue(byte v) {
        for (Type1 c : values()) {
            if (c.value == v) {
                return c;
            }
        }
        throw new IllegalArgumentException(String.valueOf(v));
    }
}
```

In such a case, an exception flies and because of this it is not possible to use micronaut-openapi normally

PR #939, opened 2023-02-22
Fix process JsonValue method when it returns another enum

altro3:


altro3:
@graemerocher Could you merge this PR and #937 and release new version? I waited a few days. So far, there are no new bugs. But this bug is unpleasant for me, because the generation of the auto-generated code using xjc is broken now in this case

altro3:
@graemerocher ping

PR #950, opened 2023-03-08
Add protection of duplicate parameters

### LINKED ISSUE #948: Duplicate path parameters when using RequestBean in a controller that inherits from base class
### Expected Behavior

If:

1. A path parameter is caught using a RequestBean
2. The RequestBean variable is not named the same as the parameter
3. The containing Controller inherits from ANY class
4. Using Kotlin (unsure about Java)

We expect the path parameter to be output only once in the openapi spec:
```openapi
/{path}/simple:
    get:
      operationId: getPath_1
      parameters:
      - name: path
        in: path
        required: true
        schema:
          type: string
      responses:
        "200":
          description: getPath_1 200 response
          content:
            application/json:
              schema:
                type: string
```

### Actual Behaviour

The parameter is output twice:
```openapi
/{path}/simple:
    get:
      operationId: getPath_1
      parameters:
      - name: path
        in: path
        required: true
        schema:
          type: string
      - name: path
        in: path
        required: true
        schema:
          type: string
      responses:
        "200":
          description: getPath_1 200 response
          content:
            application/json:
              schema:
                type: string
```

### Steps To Reproduce

1. Catch a  path parameter using a RequestBean
2. RequestBean variable is not named the same as the parameter
3. The containing Controller inherits from ANY class
4. Using Kotlin

### Environment Information

MacOS
Java 11

(For all details, see the project)

### Example Application

https://github.com/patimen/duplicatepathparams

### Version

3.7.3

altro3:
Fixes #948

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
@@ -657,7 +657,7 @@ public abstract class AbstractOpenApiEndpointVisitor extends AbstractOpenApiVisi
                         }
                         unwrappedParameter.setName(entry.getKey());
                         unwrappedParameter.setSchema(entry.getValue());
-                        swaggerParameters.add(unwrappedParameter);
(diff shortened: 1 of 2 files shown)


graemerocher:
is there a test for this one?

PR #968, opened 2023-03-25
Add support requiredMode

### LINKED ISSUE #962: Add support for the new requiredMode to replace required
## Problem

Swagger Core deprecated the `required` field in `Schema` with `requiredMode`. For more information see https://github.com/swagger-api/swagger-core/pull/4286 . Upgrading to the latest Micronaut OpenAPI + Swagger OpenAPI core version marks this field as deprecated:

Bug11_1

Switching to the new annotation value makes the compiler happy (No warning):

Bug11_2

But this new annotation value isn't picked up by Micronaut OpenAPI, so the `required` field in the generated OpenAPI file is dropped, which leads to a compiler error in our TypeScript frontend (The value is now undefined):

Bug11_3

## Solution

Is it possible to add support for `requiredMode` which takes precedence over `required`? There was another annotation value that was deprecated but I can't remember.

## Example

You can reproduce the problem with bug 11 in https://github.com/swaechter/openapiproblems

# Version

* Platform: Linux Debian
* Java: OpenJDK 64-Bit Server VM (build 17.0.4+8-Debian-1deb11u1, mixed mode, sharing)
* Micronaut Application Gradle Plugin: 3.7.2
* Micronaut Version: 3.8.7

altro3:
Fixed #962

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -1005,10 +1005,19 @@ abstract class AbstractOpenApiVisitor {
                 // check schema required flag
                 AnnotationValue<io.swagger.v3.oas.annotations.media.Schema> schemaAnnotationValue = element.getAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
                 Optional<Boolean> elementSchemaRequired = Optional.empty();
+                boolean isAutoRequiredMode = true;
                 boolean isRequiredDefaultValueSet = false;
                 if (schemaAnnotationValue != null) {
                     elementSchemaRequired = schemaAnnotationValue.get("required", Argument.of(Boolean.TYPE));
                     isRequiredDefaultValueSet = !schemaAnnotationValue.contains("required");
+                    io.swagger.v3.oas.annotations.media.Schema.RequiredMode requiredMode = schemaAnnotationValue.enumValue("requiredMode", io.swagger.v3.oas.annotations.media.Schema.RequiredMode.class).orElse(null);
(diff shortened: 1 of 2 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -1005,10 +1005,19 @@ protected void processSchemaProperty(VisitorContext context, TypedElement elemen
                 // check schema required flag
                 AnnotationValue<io.swagger.v3.oas.annotations.media.Schema> schemaAnnotationValue = element.getAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
                 Optional<Boolean> elementSchemaRequired = Optional.empty();
+                boolean isAutoRequiredMode = true;
                 boolean isRequiredDefaultValueSet = false;
                 if (schemaAnnotationValue != null) {
                     elementSchemaRequired = schemaAnnotationValue.get("required", Argument.of(Boolean.TYPE));
                     isRequiredDefaultValueSet = !schemaAnnotationValue.contains("required");
+                    io.swagger.v3.oas.annotations.media.Schema.RequiredMode requiredMode = schemaAnnotationValue.get("requiredMode", io.swagger.v3.oas.annotations.media.Schema.RequiredMode.class).orElse(null);
```
graemerocher, line 1013: use `enumValue(..)` instead of converting in this way it is less expensive

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
altro3: fixed

PR #969, opened 2023-03-25
Fix enum schema type

### LINKED ISSUE #963: Enum schema names are nested and malformed
### Expected Behavior

The upgrade from Micronaut 3.7.7 to Micronaut 3.8.0 shouldn't break the OpenAPI

### Actual Behavior

After upgrading from Micronaut 3.7.7 to Micronaut 3.8.0, more or less all enum schemas in our application/OpenAPI files are nested to another schema and thus malformed (Prefixed with `Enum_` and with a trailing underscore). Example DTO with several enum/sub classes:

Screenshot1

The generated OpenAPI references these subschemas as expected:

Screenshot2

But following these enum schemas leads to a second, nested schema that is named differently to prevent schemas with the same name:

Screenshot3

Screenshot4

In addition, in the `BackupSettingsDto`, the value of the enum field now have an added `object` type (In 3.7.7 it was just a `allOf + $ref`)

### Steps To Reproduce

That is the point that confuses me: I tried to reproduce the problem in https://github.com/swaechter/openapiproblems/tree/master/src/main/java/ch/swaechter/bug11 , but failed to do so. I am not sure what is the difference between my project and the bug example repo. I hope @altro3 can help out (There were several enum/JsonValue changes from reading the changelogs).

### Environment Information

* Platform: Linux Debian
* Java: OpenJDK 64-Bit Server VM (build 17.0.4+8-Debian-1deb11u1, mixed mode, sharing)
* Micronaut Application Gradle Plugin: 3.7.2
* Micronaut Version: 3.8.0

### Example Application

Not reproducible at the moment. https://github.com/swaechter/openapiproblems/tree/master/src/main/java/ch/swaechter/bug11 might serve as a starting point.

### Version

3.8.0

altro3:
Fixed #963

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/ConvertUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/ConvertUtils.java
@@ -40,11 +40,20 @@ import java.util.UUID;
 import javax.xml.datatype.XMLGregorianCalendar;
 
 import io.micronaut.core.annotation.AnnotationValue;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.beans.BeanMap;
+import io.micronaut.core.util.CollectionUtils;
+import io.micronaut.inject.ast.ClassElement;
+import io.micronaut.inject.ast.ElementQuery;
+import io.micronaut.inject.ast.EnumElement;
+import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.visitor.VisitorContext;
 import io.micronaut.openapi.swagger.ObjectMapperFactory;
+import io.micronaut.openapi.swagger.PrimitiveType;
 import io.swagger.v3.oas.models.security.SecurityRequirement;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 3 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/ConvertUtils.java
```diff
@@ -188,7 +196,36 @@ public static SecurityRequirement mapToSecurityRequirement(AnnotationValue<io.sw
...
+     * @return pair with openapi type and format
+     */
+    @NonNull
+    public static Pair<String, String> checkEnumJsonValueType(@NonNull EnumElement type, @Nullable String schemaType, @Nullable String schemaFormat) {
+        if (schemaType != null && !schemaType.equals(PrimitiveType.STRING.getCommonName())) {
+            return Pair.of(schemaType, schemaFormat);
+        }
+        Pair<String, String> result = null;
+        // check JsonValue method
+        for (MethodElement method : type.getEnclosedElements(ElementQuery.ALL_METHODS)) {
```
graemerocher, line 214: this is very expensive, use `getEnclosedElements(ElementQuery.ALL_METHODS).annotated(metadata -> metadata.isAnnotationPresent(JsonValue.class)` to reduce the cost of this

graemerocher (changes requested):

Issue #964, opened 2023-03-23
Empty properties when using `@Table` annotation and Groovy subprojects


tcrespog:
### Expected Behavior

The properties in the generated spec of a response class are properly populated when the class of one of its attributes is annotated with `javax.persistence.Table` regardless of the project structure (subprojects) and the language (Groovy).

### Actual Behaviour

- Given a project made of two Groovy subprojects.
- And, in the first subproject, two classes: 
   - `Pet`: JPA `@Entity` class.
   - `GetPetResponse`: wrapper of the `Pet` class.
- And, in the second subproject, a single `@Controller` class and method that returns a `GetPetResponse` object as response body.
- When the `Pet` class contains the `@Table` annotation, the generated spec of `GetPetResponse` does not include the expected properties.

Any of the following actions make it work:
- Removing the `@Table` annotation.
- Moving the `GetPetResponse` class to the subproject where the `@Controller` class is located.
- Using Java instead of Groovy.  

### Steps To Reproduce

1. Download the `mn-openapi-table-issue` sample project.
2. Build the `app` subproject and check the generated OpenAPI YML spec (at `app/build/classes/groovy/main/META-INF/swagger`): the `properties` section of `GetPetResponse` is empty.
3. Remove the `@Table` annotation in `Pet`.
4. Re-build the `controllers` subproject and check the generated OpenAPI YML spec: the `properties` section of `GetPetResponse` is properly populated.

### Environment Information

JDK version: 17
Language: Groovy

### Example Application

https://github.com/tcrespog/mn-openapi-table-issue

### Version

3.8.7

altro3:
@tcrespog Hi! It turned out to be one of the hardest bugs to reproduce :-). After several hours of analysis, I finally found the problem.

So, the problem arises because of a bug in the `io.micronaut.data:micronaut-data-hibernate-jpa` library (and the `io.micronaut.sql:micronaut-hibernate-jpa` library). The problem occurs when processing the `@Table` annotation and only when working with groovy. Unexpected behavior when processing the `indexes` annotation attribute. The problem only occurs if this field is an empty array.

I will fix the bug in these libraries and everything will be fine in the next release.

In order not to wait for a new release, I came up with a quick fix for you: just add one Index annotation to the Table annotation. Like that:
```groovy
@Table(name = 'petstore_pet', indexes = @Index)
```

altro3:
[...]

PR #970, opened 2023-03-26
Add logging, if we have problems with getting beanProperties

altro3:
See #964


altro3:
@graemerocher ping

graemerocher:
not sure about this one, in general you shouldn't use System.out but the `VisitorContext.warn` method

altro3:
@graemerocher oh, yes, I fix it now

altro3:
@graemerocher Stop. And I don't use `System.out`. I use exactly `context.warn` . What's wrong?

graemerocher:
sorry misread

PR #978, opened 2023-04-04
Add ability to disable micronaut-openapi

### LINKED ISSUE #977: Add option to disable(/enable) OpenAPI generation
## User Story

When running the Micronaut development server via `./gradlew clean run -t`, OpenAPI and views are generated, which takes extra time. I want to minimize the time taken for the server to start so that Gradle's restart on change provides the best DX / productivity.

## Proposed Solution

This should be possible in some way by providing settings depending on the Gradle task. That is, for the `run` task, provide settings to disable OpenAPI and views.

Settings for enabling/disabling views seem to exist. e.g. the system properties equivalent of `redoc.enabled=true`.
See https://micronaut-projects.github.io/micronaut-openapi/latest/guide/#viewsGenerationWithSystemProperties

A setting for controlling OpenAPI generation does not seem to exist. I propose adding it. Mirror views settings by making it an `enabled` setting. Set the default to true so that it is not a breaking change.

## Alternative Considerations

Make it a '`disableOpenApi`' setting. Default to false. Allow users to set to true.

## After the fact alternative consideration

Move OpenAPI to its own Gradle task (would involve adding a Gradle plugin?)

altro3:
Fixed #977

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
@@ -163,6 +163,9 @@ public abstract class AbstractOpenApiEndpointVisitor extends AbstractOpenApiVisi
      * @param context The visitor context
      */
     public void visitClass(ClassElement element, VisitorContext context) {
+        if (!Utils.isOpenApiEnabled()) {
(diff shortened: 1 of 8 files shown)


graemerocher:
@altro3 could you resolve the conflicts?

altro3:
@graemerocher done

PR #982, opened 2023-04-14
fixed single character issue in customSerializers.adoc

ajohnson1200:
Super small typo on line 15.

Issue #974, opened 2023-03-30
Enum string with special character not quoted in yaml


StefanSrf:
### Expected Behavior

Enum-strings with special characters should be quoted in yaml.

Example:
```
enum class DemoEnum {
    @JsonProperty("escaping-dash")
    ESCAPING_DASH,
    @JsonProperty("noEscaping")
    NO_ESCAPING
}
```
should result in 
```
    DemoEnum:
      type: string
      enum:
      - 'escaping-dash'
      - noEscaping
```

### Actual Behaviour

Invalid string in yaml, e.g. escaping-dash instead of 'escaping-dash'
```
    DemoEnum:
      type: string
      enum:
      - escaping-dash
      - noEscaping
```

### Steps To Reproduce

see example in Expected Behavior

### Environment Information

- Ubuntu
- JDK 11

### Version

3.8.7

altro3:
@StefanSrf don't need to escape it. This is normal yaml file. You can check it editor.swagger.io
![изображение](https://user-images.githubusercontent.com/48474279/228977222-aa4c9c23-f445-4990-bee7-002a248c3026.png)

Issue #944, opened 2023-03-03
Loading of Environment properties should be in sync with the running application.


francesco-monorchio:
### Feature description

This project is very awesome but i don't like so much the loading of the Environment at compile time.
Groovy has a lot of problems with it. For example it prints a lot of debug log even when "micronaut.environment.enabled" is disabled. And another problem is that the generated schema sometimes is not in sync with the running application due to the fixed environments in openapi.properties during compilation that can be different from that in MICRONAUT_ENVIRONMENTS at runtime.

In my project, I have resolved some of the problems, using a sort of placeholders during compile time generation of the schema and exposing the yml or json specification file through an endpoint that before returning it do placeholder substitution based on the current environment.

I did a simple example repo to demonstrate this process here (https://github.com/francesco-monorchio/micronaut-openapi-example). WDYT?

Thanks,
Frank

altro3:
@francesco-monorchio Hi! Unfortunately, these are limitations of the annotation processor. It always runs BEFORE compiling the project. This allows you not to add additional dependencies at the start of the application and separate business logic and auxiliary logic.

The only solution that I see is an additional module for processing placeholders at application start, but it looks like a crutch.

francesco-monorchio:
I know that Groovy has problems with annotation processors :-( . But, instead, what about the placeholder substituion at runtime?
Why do you think it is a crutch? In that way we can dinamically update the schema based on the current active environments with no change and especially no rebuild. Have you looked at my implementation? It's not perfect for sure, but for me, can be a good improvement. :-)

altro3:
Actually, no one forbids you to substitute placeholders in swagger.yaml at the start of the application. Wait, you need to write a custom handler for this and generate the final swagger.yaml when the application starts. The annotation processor cannot physically do this, because it does not participate in the launch of the final application and generates the swagger file before compilation.

That's why I say that the only solution is to write a separate code that will be executed when the application starts to generate a swagger file with substituted placeholders. But I see no reason to strain so much in such a simple and particular case.

My vision is this: you can write the code yourself that loads the generated swagger file, substitutes placeholders and creates the final swagger file, such as you need

PR #986, opened 2023-04-15
Minor fixes for allOf and other improvements.

altro3:
- Changed `context.fail` to `context.warn` when path variable is not found (!critical because now if pathVariable is not found the build will fail)
 - Fix duplicates references in allOf block
 - Sorting schemas in allOf block: schemas with ref must be first

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ projectVersion=4.8.7-SNAPSHOT
 projectGroup=io.micronaut.openapi
 
 micronautDocsVersion=2.0.0
-micronautVersion=3.8.8
+micronautVersion=3.8.9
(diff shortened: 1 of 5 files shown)


altro3:
@graemerocher Hi! Could you merge this PR and #982 and finalize release?

Also, please, close next tickets:

Fixed: #977, #104

Won't do: #944, #904

Not a bug: #974

graemerocher:
@altro3 as a contributor you should be able to close those tickets

altro3:
@graemerocher Hi! Thank you for your trust! Yes, now I can do it myself :-). Please don't create a new release just yet. I found a few more small inaccuracies, I would like to fix them before the release is finalized

PR #1010, opened 2023-05-06
Fixed @ApiResponse annotations processing

### LINKED ISSUE #1008: @ApiResponse annotations are only seeked on declared annotations or stereotype instead of all annotations or stereotypes
### Expected Behavior

For the sake of visibility, I use to put Swagger annotations on an interface instead of the implementing controller.

Annotations should be processed in this case.

It works as expected in micronaut-openapi prior to 4.5.0. [This commit](https://github.com/micronaut-projects/micronaut-openapi/commit/e4029a7974c58286d541d2b93076d65fd223d9c7#diff-9ab68a59a7bb42a88c3f39e528a533d7cf9da75188c0ba1551d74b4131181be3L1240) introduces the issue.

Before we had `element.getAnnotationValuesByType(io.swagger.v3.oas.annotations.responses.ApiResponse.class);` but after it was `element.getDeclaredAnnotationValuesByType(io.swagger.v3.oas.annotations.responses.ApiResponse.class);`

### Actual Behaviour

@ApiResponse annotations are not processed if declared on an interface, only on the implementing controller

### Steps To Reproduce

- Write a controller and an interface (the controller implements the interface), and put 2 (at least) @ApiResponse on an interface method
- Compile and inspect openapi yaml file
- The ApiResponse are not processed

### Environment Information

_No response_

### Example Application

https://github.com/ggerbaud/demo-openapi-apiresponse

### Version

micronaut-openapi 4.5.0+

altro3:
Fixed #1008


openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
```diff
@@ -1516,6 +1514,10 @@ private void processResponses(io.swagger.v3.oas.models.Operation operation, List
         }
         if (CollectionUtils.isNotEmpty(responseAnnotations)) {
             for (AnnotationValue<io.swagger.v3.oas.annotations.responses.ApiResponse> response : responseAnnotations) {
+                String responseCode = response.get("responseCode", String.class).orElse("default");
```
dstepanov, line 1517: Use `stringValue("responseCode")`

dstepanov (changes requested):

dstepanov (approved):

PR #1011, opened 2023-05-06
Fixed duplicate elements inside oneof and anyof blocks

altro3:


altro3:
These changes included here : #1012

PR #1014, opened 2023-05-10
Fixed process kotlin coroutines and reactivex types

### LINKED ISSUE #1013: Correctly read coroutine Flow
### Feature description

Hi! 
I'm trying to build api using coroutines and on endpoint where i respondd with flow I get `Flow_ServiceProviderResponse_` without any schema.
I guess it should be added here https://github.com/micronaut-projects/micronaut-openapi/blob/master/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java#L1268

P.S. Openapi don't work with KSP, any chance to fix it in the future?

altro3:
Fixed #1013


openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
```diff
@@ -0,0 +1,144 @@
...
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.inject.ast.ClassElement;
+import io.micronaut.inject.ast.TypedElement;
+
+/**
+ * Some util methods.
+ *
+ * @since 4.8.3
+ */
+public final class ElementUtils {
```
graemerocher, line 35: annotate with `@Internal`

openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
```diff
@@ -0,0 +1,144 @@
...
+            || element.hasStereotype("javax.annotation.Nullable")
+            || element.hasStereotype("jakarta.annotation.Nullable")
+            || element.hasStereotype("org.jetbrains.annotations.Nullable")
+            || element.hasStereotype("androidx.annotation.Nullable")
+            || element.hasStereotype("edu.umd.cs.findbugs.annotations.Nullable")
+            || element.hasStereotype("org.eclipse.jdt.annotation.Nullable")
+            || element.hasStereotype("io.reactivex.annotations.Nullable")
+            || element.hasStereotype("io.reactivex.rxjava3.annotations.Nullable")
+            || element.hasStereotype("reactor.util.annotation.Nullable")
+            || element.hasStereotype("org.jspecify.annotations.Nullable");
```
graemerocher, line 87: hard coding this is not great, these should be replaced by classes like this:

https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/core-processor/src/main/java/io/micronaut/inject/annotation/internal/JakartaNullableTransformer.java

Perhaps in core

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
altro3: @graemerocher By the way, how is this solution worse than what you suggest? You propose to hardcode the names of annotations separated by classes, I have them in one place, without using additional complex structures and the check goes linearly. In both cases, you will have to hardcode the full path to the annotation class, only in your case I will have to describe 10 more classes without any profit.

In general, I do not understand the better your solution :-)

openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @return true if element is nullable, false - otherwise.
+     */
+    public static boolean isNullable(TypedElement element) {
+        return element.isNullable()
+            || element.getType().isOptional()
+            || element.hasStereotype(Nullable.class)
+            || element.hasStereotype("javax.annotation.Nullable")
+            || element.hasStereotype("jakarta.annotation.Nullable")
+            || element.hasStereotype("org.jetbrains.annotations.Nullable")
```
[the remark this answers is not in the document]
altro3: @graemerocher Hah! As you see, you're not right ;-)
изображение
Reverted

openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
graemerocher: because if it is part of core then every module does not have to repeat the same logic and it is encapsulated into a single place. Other parts of the framework will likely need to know what is nullable and that could differ from what OpenAPI believes is nullable if it is only captured here.

altro3: @graemerocher do you suggest to add these annotaions to core module?

graemerocher: that is what makes sense indeed

altro3:
@graemerocher Hi! Could you approve this PR and this: https://github.com/micronaut-projects/micronaut-openapi/pull/988 . After it we can finalize release 4.9.0

altro3:
@graemerocher ping

altro3:
@graemerocher Thanks! Can I publish the release myself or will you do it yourself?? This is draft release: https://github.com/micronaut-projects/micronaut-openapi/releases/tag/untagged-6a3ba8a3c25774f62800

altro3:
@graemerocher And about merge to master branch: it's to early. First need merge micronaut 3.9.2 changes to micronaut 4.0.0 (feature to disable slf4j).

PR #1040, opened 2023-06-02
Initial re-integration of the OpenAPI generator into Micronaut OpenAPI

melix:
This commit migrates the OpenAPI code generator from the external OpenAPI generator project to this project. A new module named `openapi-generator` depends on OpenAPI generator and replicates features which are currently found in that project. However, there are a few changes:

- the build file generation has been removed. Because we're going to integrate this in build tools, it doesn't make sense to generate these. Even if we use it in Micronaut Launch, then it's the responsibility of Launch to generate the files. One drawback is that the user has to figure out which dependencies to add for the client or server, if they use the build tool plugins.
- a type-safe API has been introduced to configure options. It is actually hard to figure out all the options we can configure and are relevant to the code, this work is therefore in progress.
- the default package for javax annotations was moved to jakarta

Test projects are used to make sure that we can generate at least one client and one server, with hardcoded options. We're using generated sources, which are not supposed to be changed by the user. It simulates what the build tool plugins would do, except that these would probably generate sources in different folders based on what we generate.

@andriy-dmytruk can you figure out why the test `io.micronaut.openapi.generator.MicronautServerCodegenTest#testExtraAnnotations` is failing? Apparently an extra annotation declared in the spec isn't applied to a field. It may be related to renames I have made but I am not sure.


graemerocher (approved):

sdelamo:
we missed documentation in this PR.

PR #1054, opened 2023-06-10
Fix controller inheritance processing

altro3:


graemerocher:
is there a test for this?

graemerocher:
or at least an issue report describing what it fixes?

altro3:
Not sure if I can write a test for this case. I was only able to find the problem while debugging the code in my project. So, the problem arose in the case when the parent controller had the Controller annotation, and the child class did not have the Controller annotation. And one more prerequisite (why I didn’t write a test): the problem arose only if the parent controller was described in one of the jar files!

The problem is that when I wrote this one, I assumed that AbstractOpenApiEndpointVisitor.visitClass was always called first and I removed CONTEXT_CHILD_PATH from the context AT THE BEGINNING of the method.

However, as it turned out, in my case, since the annotation was described in the parent class and in the jar file, the AbstractOpenApiEndpointVisitor.visitClass method was not called, but the AbstractOpenApiEndpointVisitor.visitMethod method was immediately called.

Thus, the CONTEXT_CHILD_PATH variable was not removed from the context and therefore incorrect paths were obtained in this case.

Very unusual problem, spent a lot of time. to understand what the problem is.

For example:

Parent controler class (must be in jar file)
```java
@Controller("/parentpath")
class ParentController {

    @Post("/postMethod")
    void post() {}
}
```
Child controller class (must be without Controller annotation)
```java
class ChildController extends ParentController{
}
```

altro3:
@graemerocher Could you merge these 3 fixes and finalize release 4.9.1 ?

altro3:
or I can do it myself?

altro3:
@graemerocher Could you finalize release 4.9.1? 

I created draft here: https://github.com/micronaut-projects/micronaut-openapi/releases/tag/v4.9.1

Or I can do it myself too?

PR #1055, opened 2023-06-11
Fix processing multipart form data

altro3:


openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
```diff
@@ -89,6 +89,21 @@ public static boolean isNullable(TypedElement element) {
...
+    /**
+     * Checking if the type is file upload type.
+     *
+     * @param type type element
+     *
+     * @return true if this type one of known file upload types
+     */
+    public static boolean isFileUpload(ClassElement type) {
+        String typeName = type.getName();
+        return "io.micronaut.http.multipart.StreamingFileUpload".equals(typeName)
```
graemerocher, line 101: are these types not accessible in the context?

PR #1077, opened 2023-06-21
Upgrade to core 4.0.0-RC3

melix:


sdelamo (approved):
Why did you change to use the Micronaut HTTP Client Java based instead of the net one?

melix:
> Why did you change to use the Micronaut HTTP Client Java based instead of the net one?

mmm, I simply changed the implementation to use `micronaut-http` because the generated code uses annotations from this project. I moved the `micronaut-http-client` to `testImplementation` only because it wasn't required by the generated server code but only for tests.

PR #1089, opened 2023-06-28
Add parameter and response body mapping

andriy-dmytruk:
These changes are to allow the use of custom types for parameters and response bodies to utilize all Micronaut capabilities for code simplification.

## Parameters mapping

As an example, if we add the following parameter mappings:
```js
[
  {name: "page", location: "QUERY", mappedType: "io.micronaut.data.model.Pageable"},
  {name: "size", location: "QUERY", mappedType: "io.micronaut.data.model.Pageable"},
  {name: "sortOrder", location: "QUERY", mappedType: "io.micronaut.data.model.Pageable"},
]
```
micronaut `Pageable` will be used for `page`, `size` or `sortOrder` headers:
```java
    @Get("/sendPageQuery")
    @Produces({"application/json"})
    Mono<String> sendPageQuery(
        @NotNull
        Pageable pageable
    );
```
instead of each header being a separate parameter:
```java
    @Get("/sendPageQuery")
    @Produces({"application/json"})
    Mono<String> sendPageQuery(
        @HeaderParam
        Integer page,
        @HeaderParam
        Integer size,
        @Headerparam
        String sortOrder
    );
```

The spec in this case is :
```yaml
operations:
  /sendPageQuery:
    get:
      operationId: sendPageQuery
      tags: [ parameters ]
      description: A method that takes page query as its argument
      parameters:
        - $ref: '#/parameters/PageQueryParam'
        - $ref: '#/parameters/PageSizeQueryParam'
        - $ref: '#/parameters/PageSortQueryParam'
      responses:
        200:
          description: Success
          schema:
            type: string
parameters:
  PageQueryParam:
    name: page
    in: query
    type: integer
    minimum: 0
    default: 0
    description: The page number to retrieve starting from 0.
  PageSizeQueryParam:
    name: size
    in: query
    type: integer
    minimum: 1
    default: 10
    description: The number of items per page.
  PageSortQueryParam:
    name: sortOrder
    in: query
    type: string
    description: |
      Parameter describing the sort. Allows specifying the sorting direction using the keywords {@code asc} and
      {@code desc} after each property. For example, {@code "sort=name,desc,&sort=age"} will sort by name in descending
      order and age in ascending.
```

Similarly to the example above, custom types can be created with `TypedRequestArgumentBinder` implementations. Parameters can be ignored by setting the `mappedType` to `null`, as some parameters may be consumed inside interceptors or filters. 

## Response body mapping

The body mapping works in a similar way, but changes the return type of operation. Since server can only return header parameters, the mapping is done based on them.

For example, if we add the following body mapping:
```js
{headerName: "Last-Modified", mappedBodyType: "io.micronaut.openapi.test.dated.DatedResponse"}
```
the operation would be created as (if a `Last-Modified` header is specified):
```
    @Get("/getDatedSimpleModel")
    @Produces({"application/json"})
    Mono<DatedResponse<SimpleModel>> getDatedSimpleModel();
```
The custom `DatedResponse` object then would need a `MessageBodyWriter` implementation and include the required header.

The specification in the case above:
```yaml
operations:
  /getDatedSimpleModel:
    get:
      operationId: getDatedSimpleModel
      tags: [ responseBody ]
      description: A method to get a simple model with last-modified header
      responses:
        200:
          description: Success
          headers:
            Last-Modified: *Last-Modified-header
          schema:
            $ref: '#/definitions/SimpleModel'
x-headers:
  LastModifiedHeader: &Last-Modified-header
    name: Last-Modified
    type: string
    format: date-time
    description: The last time an entity returned with the response was modified.
```

## Note
We can make this interface available in gradle/maven plugins or keep available for extension generators only.

### DIFF
--- a/test-suite-generator-util/src/main/java/io/micronaut/openapi/testsuite/GeneratorMain.java
+++ b/test-suite-generator-util/src/main/java/io/micronaut/openapi/testsuite/GeneratorMain.java
@@ -108,8 +108,6 @@ public class GeneratorMain {
     }
 
     private static List<Map<String, String>> parseListOfMaps(String string) {
-        System.out.println("json: " + string);
-
         List<Map<String, String>> result = new ArrayList<>();
         if (string.isBlank()) {
             return result;
--- a/test-suite-generator-util/src/main/java/io/micronaut/openapi/testsuite/GeneratorMain.java
+++ b/test-suite-generator-util/src/main/java/io/micronaut/openapi/testsuite/GeneratorMain.java
@@ -15,12 +15,18 @@
  */
 package io.micronaut.openapi.testsuite;
 
+import io.micronaut.openapi.generator.AbstractMicronautJavaCodegen;
 import io.micronaut.openapi.generator.MicronautCodeGeneratorEntryPoint;
 
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * An entry point to be used in tests, to simulate
@@ -42,6 +48,7 @@ public class GeneratorMain {
      */
     public static void main(String[] args) throws URISyntaxException {
         boolean server = "server".equals(args[0]);
+        List<AbstractMicronautJavaCodegen.ParameterMapping> parameterMappings = parseParameterMappings(args[4]);
         MicronautCodeGeneratorEntryPoint.OutputKind[] outputKinds
             = Arrays.stream(args[3].split(","))
             .map(MicronautCodeGeneratorEntryPoint.OutputKind::of)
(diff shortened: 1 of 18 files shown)


test-suite-generator-util/src/main/java/io/micronaut/openapi/testsuite/GeneratorMain.java
```diff
@@ -74,4 +87,73 @@ public static void main(String[] args) throws URISyntaxException {
...
+        return parseListOfMaps(string).stream().map(map -> new AbstractMicronautJavaCodegen.ResponseBodyMapping(
+            map.get("headerName"),
+            map.get("mappedBodyType"),
+            "true".equals(map.get("isListWrapper")),
+            "true".equals(map.get("isValidated"))
+        )).collect(Collectors.toList());
+    }
+
+    private static List<Map<String, String>> parseListOfMaps(String string) {
+        System.out.println("json: " + string);
```
yawkat, line 111:
```suggestion
```

yawkat (approved):

melix (approved):

graemerocher (approved):

PR #1098, opened 2023-07-02
OpenAPI Generator Return type Improvements

andriy-dmytruk:
Improvements for the operation return type:
* `HttpResponse` can be generated only where required (a response header is used or status code other than 200)
* `Flux` is returned for iterable types instead of `Mono<List<>>` when using reactive return types


graemerocher:
there are failing tests

graemerocher:
`Flux` might be worse for performance /cc @dstepanov

dstepanov:
Flux or publisher response is always streaming the content. In my benchmarking Flux<Book> is much slower than Mono<List<Book>>

If you have a single response use Mono, or @SingleResult Publisher / Flux. For returning a list of values is always better to use collectList and Mono<List>

PR #1112, opened 2023-07-13
Add support for @JsonView

### LINKED ISSUE #797: Add support for @JsonView
### Feature description

It appears there is no support for `@JsonView` annotation. This was added to swagger-core in https://github.com/swagger-api/swagger-core/pull/2662 

I'd like to be able to exclude certain fields from the response body schema based on the `@JsonView` configuration but they get added to the schema regardless of the annotation now so both of my responses with the same object, but different views have the exact same schema.

altro3:
Fixes #797

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
@@ -65,6 +65,7 @@ import io.micronaut.core.type.Argument;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
+import io.micronaut.http.HttpStatus;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.uri.UriMatchTemplate;
 import io.micronaut.http.uri.UriMatchVariable;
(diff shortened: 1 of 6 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
```diff
@@ -158,4 +167,33 @@ private static boolean findAnyAssignable(ClassElement type, List<String> typeNam
...
+        AnnotationClassValue[] test = (AnnotationClassValue[]) values.get("value");
+        if (ArrayUtils.isEmpty(test)) {
+            return null;
+        }
+        List<String> jsonViewClasses = new ArrayList<>(test.length);
+        for (AnnotationClassValue value : test) {
+            jsonViewClasses.add(value.getName());
+        }
+
+        return jsonViewClasses;
```
graemerocher, line 197: I believe you can just call `String[] classNames = element.getAnnotationMetadata().stringValues()`

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
```diff
@@ -2388,6 +2446,41 @@ private void processPropertyElements(OpenAPI openAPI, VisitorContext context, El
...
+                }
+            }
+        } else {
+            ClassElement superType = jsonViewClassEl.getSuperType().orElse(null);
+            if (superType != null) {
+                return checkJsonView(fieldJsonViewClasses, superType);
+            }
+        }
+
+        return false;
```
graemerocher, line 2481: You can probably just call `visitContext.getClassElement(fieldJsonViewClassName)` to get the JsonViewClass and then just use `isAssignable` instead of manually traversing the class hierarchy? Seems like you have just reimplemented an assignment check

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
altro3: fixed

graemerocher (approved):

PR #1115, opened 2023-07-14
Add ability to set project directory.

### LINKED ISSUE #1101: kapt(mn.micronaut.openapi) breaks incremental annotation processing
### Expected Behavior

An addition of `kapt(mn.micronaut.openapi)` to the gradle build configuration doesn't lead to extreme changes in incremental build time

### Actual Behaviour

An addition of `kapt(mn.micronaut.openapi)`
leads to extremely slow rebuild time with any small change, which is disruptive for the execution of single tests from IDE. 
The following warnings are new during the build process:

```
> Task :kaptKotlin
[WARN] Issue detected with io.micronaut.annotation.processing.TypeElementVisitorProcessor. Expected 1 originating source file when generating <>/build/generated/source/kapt/main/dummy12600902257963, but detected 0: [].
Note: Generating OpenAPI Documentation
...
Note: Writing OpenAPI file to destination: <>/build/tmp/kapt3/classes/main/META-INF/swagger/<>-2.0.yml
Note: Writing OpenAPI views to destination: <>/build/tmp/kapt3/classes/main/META-INF/swagger/views
[WARN] Incremental annotation processing requested, but support is disabled because the following processors are not incremental: io.micronaut.annotation.processing.TypeElementVisitorProcessor (DYNAMIC).
```

### Version

micronaut-bom 3.9.4, which brings micronaut-openapi:3.8.7
Also reproduces with an upgrade to micronaut-openapi:4.9.2

### Additional context

[Potentially relevant issue](https://github.com/micronaut-projects/micronaut-core/issues/9148)

altro3:
Fixed #1101

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
@@ -135,6 +135,7 @@ import static io.swagger.v3.oas.models.Components.COMPONENTS_SCHEMAS_REF;
     OpenApiApplicationVisitor.MICRONAUT_OPENAPI_SECURITY_ENABLED,
     OpenApiApplicationVisitor.MICRONAUT_OPENAPI_VERSIONING_ENABLED,
     OpenApiApplicationVisitor.MICRONAUT_OPENAPI_JSON_VIEW_DEFAULT_INCLUSION,
+    OpenApiApplicationVisitor.MICRONAUT_OPENAPI_PROJECT_DIR,
 })
 public class OpenApiApplicationVisitor extends AbstractOpenApiVisitor implements TypeElementVisitor<OpenAPIDefinition, Object> {
 
@@ -176,6 +177,14 @@ public class OpenApiApplicationVisitor extends AbstractOpenApiVisitor implements
      * System property that specifies the location of additional swagger YAML and JSON files to read from.
      */
     public static final String MICRONAUT_OPENAPI_ADDITIONAL_FILES = "micronaut.openapi.additional.files";
+    /**
+     * System property that specifies the location of current project.
+     */
+    public static final String MICRONAUT_OPENAPI_PROJECT_DIR = "micronaut.openapi.project.dir";
+    /**
+     * Loaded project directory from system properties.
+     */
+    public static final String MICRONAUT_INTERNAL_OPENAPI_PROJECT_DIR = "micronaut.internal.openapi.project.dir";
     /**
      * System property that specifies the default security schema name, if it's not specified by annotation SecurityScheme.
      */
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/Utils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/Utils.java
@@ -42,6 +42,9 @@ import io.micronaut.openapi.visitor.group.OpenApiInfo;
 import io.swagger.v3.oas.models.Components;
 import io.swagger.v3.oas.models.OpenAPI;
 
+import static io.micronaut.openapi.visitor.OpenApiApplicationVisitor.MICRONAUT_INTERNAL_OPENAPI_PROJECT_DIR;
+import static io.micronaut.openapi.visitor.OpenApiApplicationVisitor.MICRONAUT_OPENAPI_PROJECT_DIR;
+
 /**
  * Some util methods.
  *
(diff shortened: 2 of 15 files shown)


altro3:
@graemerocher Hi! Could you review and merge last 3 PR's (#1115, #1112, #1113) ? 

After it we can release 4.10.0 - last of 4.x series. Then I move these changes to relese 5.0 and all next changes will be in release 5.0 (except for bug fixes, of course)

altro3:
@graemerocher Could you also review this PR and #1113 ?

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
```diff
@@ -175,6 +176,14 @@ public class OpenApiApplicationVisitor extends AbstractOpenApiVisitor implements
...
      */
     public static final String MICRONAUT_OPENAPI_ADDITIONAL_FILES = "micronaut.openapi.additional.files";
+    /**
+     * System property that specifies the location of current project.
+     */
+    public static final String MICRONAUT_OPENAPI_PROJECT_DIR = "micronaut.openapi.project.dir";
+    /**
+     * Loaded project directory from system properties.
+     */
+    public static final String MICRONAUT_INTERNAL_OPENAPI_PROJECT_DIR = "micronaut.internal.openapi.project.dir";
```
graemerocher, line 186: Could you override `getSupportedOptions()` and provide all of these constants as annotation processor options

openapi/src/main/java/io/micronaut/openapi/visitor/Utils.java
```diff
@@ -92,14 +95,28 @@ private Utils() {
...
-        } catch (Exception e) {
-            // Should never happen
-            path = Paths.get(System.getProperty("user.dir"));
+
+        Path projectPath = context.get(MICRONAUT_INTERNAL_OPENAPI_PROJECT_DIR, Path.class).orElse(null);
+        if (projectPath != null) {
+            return projectPath;
+        }
+
+        String projectDir = System.getProperty(MICRONAUT_OPENAPI_PROJECT_DIR);
```
graemerocher, line 104: you shouldn't use system properties but the annotation processor options otherwise the compiler can't detect when input values change

graemerocher (changes requested):

altro3:
As I understand, this block do the same

изображение

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
altro3: As I understand, this block do the same

изображение

openapi/src/main/java/io/micronaut/openapi/visitor/Utils.java
altro3: I don't quite understand what's wrong here. As I understand it, it is necessary to describe the properties of the processor, which must be allowed to be loaded as system variables. They are described in the `@SupportedOptions` annotation. Now everything has been implemented.

altro3:
@graemerocher Can I merge it?

sdelamo (changes requested):
@altro3 can you add documentation about this? when should people use it and how to use it

altro3:
@sdelamo done

openapi/src/main/java/io/micronaut/openapi/visitor/Utils.java
graemerocher: in general parameters to an annotation processor should never be ready from system properties and should always be read from annotation processor arguments, this particular processor abuses that badly in several places which is less than ideal.

altro3: I see, then how is it correct to write the reading of the arguments? Although I'm not sure if there's much to worry about using system variables. I think it's redundant

altro3: @graemerocher Do I need to write something like this?
```java
context.getOptions().get(MICRONAUT_INTERNAL_OPENAPI_PROJECT_DIR)
```

altro3: It's just that I don't understand. I give kapt the annotation processor argument, but I read it with a system variable. That is, it turns out that all arguments are already available as system variables. Or does kapt set not an argument, but a system property?

I tested this solution on a test project - it works correctly.

sdelamo (approved):

altro3:
@graemerocher Can I publish releases `5.0.2` and `4.10.0` ?

openapi/src/main/java/io/micronaut/openapi/visitor/Utils.java
graemerocher: yes it needs to be `context.getOptions().get(MICRONAUT_INTERNAL_OPENAPI_PROJECT_DIR)`

PR #1149, opened 2023-07-29
Configuration improvements

altro3:
Improvements in classes structure:

1. Move configuration methods to ConfigUtils class
2. Move file methods to FileUtils class
3. Added autosave to context already loaded config value
4. Replace System.getProperty(...) to context.getOptions().get(...) for supported annotaton processor options

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/annotation/mappers/OpenAPIManagementAnnotationMapper.java
+++ b/openapi/src/main/java/io/micronaut/openapi/annotation/mappers/OpenAPIManagementAnnotationMapper.java
@@ -35,7 +35,7 @@ public class OpenAPIManagementAnnotationMapper implements TypedAnnotationMapper<
     }
 
     @Override
-    public List<AnnotationValue<?>> map(AnnotationValue<OpenAPIManagement> annotation, VisitorContext visitorContext) {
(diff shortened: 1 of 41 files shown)


altro3:
@graemerocher Hi! Now all as you want: I use context.getOpetions.get() instead System.getProperty()

openapi/src/main/java/io/micronaut/openapi/visitor/ConfigProperty.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Supported configuration properties.
+ *
+ * @since 4.10.0
+ */
+public interface ConfigProperty {
```
graemerocher, line 27: I would rename this to something less generic like `OpenApiConfigProperty` or something

graemerocher (approved):

openapi/src/main/java/io/micronaut/openapi/visitor/ConfigProperty.java
altro3: @graemerocher fixed

altro3:
@graemerocher sorry, I need to revert `context.getOptions().get(...)` to `System.getProperty(...)` because of it https://github.com/micronaut-projects/micronaut-core/issues/9671

PR #1164, opened 2023-08-10
Fix version select options in gh-pages

timyates:
I updated this project to 6.5.5 of the setting plugin and then ran

https://github.com/micronaut-projects/micronaut-core/wiki/Utilities-and-Scripts#documentation-version-select-dropdown-patching

Which patches the select dropdown for versions (copies the versions from the now fixed snapshot docs):

- 5.0.0-M10
- 5.0.0-M11
- 5.0.0-M12
- 5.0.0-M2
- 5.0.0-M3
- 5.0.0-M4
- 5.0.0-M5
- 5.0.0-M6
- 5.0.0-M7
- 5.0.0-M8
- 5.0.0-M9
- 5.0.x
- 5.1.0
- 5.1.x
- LATEST


altro3:
@timyates Great! But where is version 4.10.0? This is latest version for micronaut 3.x

altro3:
@timyates Also, I don't thik that we need to show these versions for people: 5.0.0-M10
5.0.0-M11
5.0.0-M12
5.0.0-M2
5.0.0-M3
5.0.0-M4
5.0.0-M5
5.0.0-M6
5.0.0-M7
5.0.0-M8
5.0.0-M9

timyates:
@altro3 

> But where is version 4.10.0? 

4.10.0 doesn't need patching, it's ok already https://micronaut-projects.github.io/micronaut-openapi/4.10.0/guide/index.html

The versions issue was introduced in the upgrade for Micronaut 4.0.0

> Also, I don't thik that we need to show these versions for people: 5.0.0-M10

I tend to agree, but there's a milestone cleanup which I believe is being discussed separately (so this will come later)

wetted (approved):

PR #1165, opened 2023-08-12
Fix readonly properties for generator.

### LINKED ISSUE #1160: Optional readonly field in the OpenAPI definition makes the server generator generate a field, that can't be set in the Java class
### Expected Behavior

Generated models should conform as closely as possible to the [OpenAPI specification for readOnly](https://swagger.io/docs/specification/data-models/data-types/#readonly-writeonly), but it is not possible to get this perfect, since the same models can be used for both input and output on the server. My best suggestion is:

1. `readOnly` properties should be generated by both the client and server generators the same way as optional properties (ie. properties that are not in the `required` list), since this is the only way to ensure that they can both be handled by serialization and validation.
2. One additional nice-to-have thing is, if **only** the **server generator** could add required `readOnly` properties to the generated model's constructor, ie. if they are in the `required` list. They must still be marked `@HardNullable` though, since otherwise it will not be possible for clients to send requests to the server without these properties.

### Actual Behaviour

1. Optional `readOnly` properties are unsettable in both server and client generated models.
3. Mandatory `readOnly` properties are parameters in the model constructors in both server and client generated models.

### Steps To Reproduce

**To reproduce "1. Optional `readOnly` properties are unsettable in both server and client generated models.":**

Generate a model class from this OpenAPI object definition with either generator:

```
type: object
properties:
  id:
    type: string
    readOnly: true
  value:
    type: string
required:
  - value
```

The `id` property is not settable in the model generated by either generator, since it is not part of the constructor parameters and the setter is private:

```
    public TestModel(String value) {
        this.value = value;
    }

...

    private void setId(String id) {
        this.id = id;
    }
```

**To reproduce "2. Mandatory `readOnly` properties are parameters in the model constructors in both server and client generated models.":**

2. Generate a model class from this OpenAPI object definition with either generator:

```
type: object
properties:
  id:
    type: string
    readOnly: true
  value:
    type: string
required:
  - id
  - value
```

The `id` is included in the constructor in the model generated by both generators.

### Environment Information

I believe it happens in all environments, but my environment is specifically:

MacOs
JDK 17

### Example Application

https://github.com/runenielsen/unsettablereadonly

Build it with Gradle and then inspect the generated class files `build/generated/openapi/generateServerOpenApiModels/src/main/java/com/example/openapi/server/model/BookInfo.java` and  `build/generated/openapi/generateClientOpenApiModels/src/main/java/com/example/openapi/client/model/BookInfo.java`.

The `optionalReadOnly` property can't be set in any way in either model, since the setter is private and it can't be set via the constructor either.

The `requiredReadOnly` is included in the constructor parameters in both models.

### Version

5.0.3

altro3:
Fixed #1160

### DIFF
--- a/openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
+++ b/openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
@@ -38,7 +38,9 @@
 public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#vendorExtensions.x-implements}}{{#-first}}implements {{{.}}}{{/-first}}{{^-first}}, {{{.}}}{{/-first}}{{#-last}} {{/-last}}{{/vendorExtensions.x-implements}}{
 {{/formatNoEmptyLines}}
     {{#serializableModel}}
-    private static final long serialVersionUID = 1L;
+
+    @Serial
+    private static final long serialVersionUID = {{{vendorExtensions.serialId}}}L;
     {{/serializableModel}}
 
     {{#vars}}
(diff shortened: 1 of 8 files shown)


altro3:
@andriy-dmytruk @dstepanov Could you review it and merge?

openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
```diff
@@ -113,13 +115,20 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#vendorE
     {{/vars}}
 
     {{#requiredPropertiesInConstructor}}
-    public {{classname}}({{#vendorExtensions.requiredVarsWithoutDiscriminator}}{{{datatypeWithEnum}}} {{name}}{{^-last}}, {{/-last}}{{/vendorExtensions.requiredVarsWithoutDiscriminator}}{{#vendorExtensions.areRequiredVarsAndReadOnlyVars}}, {{/vendorExtensions.areRequiredVarsAndReadOnlyVars}}{{#readOnlyVars}}{{{datatypeWithEnum}}} {{name}}{{^-last}}, {{/-last}}{{/readOnlyVars}}) {
+    public {{classname}}({{#vendorExtensions.requiredVarsWithoutDiscriminator}}{{#isReadOnly}}{{#vendorExtensions.isServer}}{{^-first}}, {{/-first}}{{{datatypeWithEnum}}} {{name}}{{/vendorExtensions.isServer}}{{/isReadOnly}}{{^isReadOnly}}{{^-first}}, {{/-first}}{{{datatypeWithEnum}}} {{name}}{{/isReadOnly}}{{/vendorExtensions.requiredVarsWithoutDiscriminator}}) {
```
andriy-dmytruk, line 118: We could make a vendor extension `isWrittable` instead to not repeat the code twice, but this is fine.

andriy-dmytruk (approved):

andriy-dmytruk:
Looks good, thanks @altro3

altro3:
@dstepanov Could you merge it?

PR #1166, opened 2023-08-12
Add ability to generate code without `@Generated` annotation and more

altro3:
- Add ability to generate code without `@Generated` annotation.
 - Add ability to generate compatible code with micronaut-aot.
 - Add ability to generate code with lombok annotations.
 - Remove generating unnecessary default constructor in POJOs.
 - Cosmetics for generated code.
 - Remove unnecessary default consumes / produces media type "application/json"
 - Removed braces in annotaions when we have only one element in collection

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
@@ -15,13 +15,28 @@
  */
 package io.micronaut.openapi.generator;
 
-import com.google.common.collect.ImmutableMap;
-import com.samskivert.mustache.Mustache;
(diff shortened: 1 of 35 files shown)


dstepanov:
@andriy-dmytruk Please review

@altro3 What's the point of generating Lombok-style beans? We can simply generate the getters/setters

altro3:
@dstepanov This is necessary for cases when you need a cleaner and shorter code. When you have more than 50 fields in your class, getters and setters and chain methods take up a lot of space. These files are less convenient to work with. I understand that the generation of getters and setters later also takes some time. But this function is useful for cases where you use a swagger file to generate a project skeleton, i.e. to do it once, and then to rule it with his hands. That's why I added options to make the code easier to use, not as generated, but as usual.

andriy-dmytruk (approved):

PR #1167, opened 2023-08-13
Fix generating model with discriminator.

### LINKED ISSUE #1163: Server generator doesn't add JsonSubTypes annotation on inheritance causing incorrect JSON to be generated
### Expected Behavior

When using a discriminator to serialize and deserialize objects with inheritance, I expect the server generator to add the `@JsonSubTypes` annotation to the correct generated model class, so the defined enum value is used in the discriminating field. For instance:

```
{"type":"DETAILED","name":"Never-ending Story","availability":"available","author":"Michael Ende","ISBN":"SOME_ISBN"}
```

"type" is the discriminating field in this example.

### Actual Behaviour

The server generator does not add the `@JsonSubTypes` annotation, and therefore the sub-class name is used as discriminating value, making the generated JSON incorrect, since the defined enum value should be used:

```
{"type":"DetailedBookInfo","name":"Never-ending Story","availability":"available","author":"Michael Ende","ISBN":"SOME_ISBN"}
```

"type" is the discriminating field in this example.

This behaviour makes the deserialization incompatible with clients, that send the correct JSON, since they will send the value `DETAILED` in `type` and the Micronaut service will expect `DetailedBookInfo`.

### Steps To Reproduce

1. Clone https://github.com/runenielsen/missingjsonsubtypes
2. Run the `testSerdeForDetailedBookInfo` test in the `MissingjsonsubtypesTest` test class
3. I have made a hacky fix in lines 64 - 82 in build.gradle.kts, that inserts the needed `@JsonSubTypes` annotation in the generated `BookInfo.java` file, so you can see, how I think it should look. Comment the fix out and run the test again to see, that it now fails.

The `@JsonSubTypes` annotation is added correctly by other generators, for instance: https://openapi-generator.tech/docs/generators/java/. So this is not a bug in the OpenAPI Generator core.

The annotation is actually also already imported in the generated class, it is just not used:

```
import com.fasterxml.jackson.annotation.JsonSubTypes;
```

### Environment Information

OS: MacOS
JDK: Java 17

But it should occur in any environment.

### Example Application

https://github.com/runenielsen/missingjsonsubtypes

### Version

4.0.3

### LINKED ISSUE #1168: Generators shouldn't add required discriminator property as a constructor parameter in generated models with inheritance
### Expected Behavior

Required discriminator properties in models with inheritance should not be part of the constructor parameters, since they are not required to be set (they are generated automatically based on the class on serialization).

Currently it is actually a problem, if they are set, since the property will then be added twice to the generated JSON on serialization due to a bug in Micronaut Serialization (reported here: https://github.com/micronaut-projects/micronaut-serialization/issues/550).

### Actual Behaviour

Required discriminator properties are added as constructor parameters for generated models with inheritance.

### Steps To Reproduce

1. Clone the project https://github.com/runenielsen/discriminatorgeneration
2. Build and inspect the generated classes to see, that the discriminator property `type` is added as a constructor parameter.
3. Run the test SerdeWithDiscriminatorTest and see the test case `testSerializationForDetailedBookInfoWithDiscriminatorSet` fail due to the discriminator property being added twice in the generated JSON, when it is set in the model. And see that the test case `testSerializationForDetailedBookInfoWithoutDiscriminatorSet` works correctly, when it is not set in the model. And lastly notice that the property is not set in the model on deserialization in the test case `testDeserializationForDetailedBookInfo`.

I'm aware that this issue is sort of in between a bug and a feature request, since it is mostly an annoyance to have a property in the constructor, that doesn't need to be set, and the main bug is actually in Micronaut Serialization. But I figured, I'd report it as a bug, since the fields in the Bug Report issue type fit better 🙂 

### Environment Information

OS: MacOS
JDK: 17

### Example Application

https://github.com/runenielsen/discriminatorgeneration

### Version

4.0.3

altro3:
Also removed discriminator from required properties constrcutor.

Fixed #1163
Fixed #1168


runenielsen:
@altro3 : Awesome, thanks a lot.

andriy-dmytruk (approved):

PR #1176, opened 2023-08-23
Fixes #1175

### LINKED ISSUE #1175: Micronaut 4.0.x does not set type:object for maps
### Expected Behavior

A map<string,string> should generate
```
   type: object
   additionalProperties:
      type: string
```
as it did with Micronaut 3.x

### Actual Behaviour

A map<string,string> generates
```
   additionalProperties:
      type: string
```

### Steps To Reproduce

Use micronaut launch and select openapi and micronaut 4.0.x.

Generate a simple dto such as
```
public class SimpleDto {
	private Map<String, String> tags;
	public SimpleDto() {}

	public Map<String, String> getTags() {
		return tags;
	}

	public void setTags(Map<String, String> tags) {
		this.tags = tags;
	}
}
```

And a simple controller like
```
@Controller("/demo")
public class DemoController {
	@Get(uri = "/test", produces = "application/json")
	public SimpleDto test() {
		return new SimpleDto();
	}
}
```

Build to generate the DTO which will be missing the `type: object`.  Following the steps with 3.x works fine.

### Environment Information

- Majaro Linux (6.1.44-1-MANJARO)
- JDK 19

### Example Application

https://github.com/mdavis95/demo/

### Version

4.0.4

andriy-dmytruk:
Solve the issue of type=object being removed for MapSchema properties.

### DIFF
--- a/openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiMapSchemaSpec.groovy
+++ b/openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiMapSchemaSpec.groovy
@@ -1,0 +1,121 @@
+package io.micronaut.openapi.visitor
+
+import io.micronaut.openapi.AbstractOpenApiTypeElementSpec
+import io.swagger.v3.oas.models.OpenAPI
+import io.swagger.v3.oas.models.Operation
+import io.swagger.v3.oas.models.media.ComposedSchema
+import io.swagger.v3.oas.models.media.MapSchema
+import io.swagger.v3.oas.models.media.Schema
+import io.swagger.v3.oas.models.media.StringSchema
+
+class OpenApiMapSchemaSpec extends AbstractOpenApiTypeElementSpec {
+
+    void "test build OpenAPI doc for map properties"() {
+
+        when:
+        buildBeanDefinition('test.MyBean', '''
+package test;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.http.HttpResponse;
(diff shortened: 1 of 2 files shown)


openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiMapSchemaSpec.groovy
```diff
@@ -0,0 +1,121 @@
...
+        petSchema.properties.size() == 2
+        petSchema.properties['map'] instanceof MapSchema
+        petSchema.properties['complexMap'] instanceof MapSchema
+
+        when:
+        MapSchema map = (MapSchema) petSchema.properties['map']
+        MapSchema complexMap = (MapSchema) petSchema.properties['complexMap']
+
+        then:
+        map.type == 'object'
```
andriy-dmytruk, line 115: For some reason this test didn't catch the issue, but when I inspected the created `OpenAPI` object in `OpenApiApplicationVisitor` ([here](https://github.com/micronaut-projects/micronaut-openapi/blob/25f05ad73d2db77722b2a6917bf576a63bcbef0a/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java#L501)) the map property had the incorrect type.

I am not sure how this schema is different from the one that is there. @altro3 do you know?

altro3: I can only watch it tomorrow. But, as the ticket author wrote, the problem has already been resolved in version 5.1.0.

I can add that the generated OpenApi (to which you put down properties throughout the entire logic) and the one that will be in the final are different objects, because this object is serialized and then deserialized using Jackson. Those. some of the properties may differ from the original object after deserialization according to the swagger rules.

altro3 (approved):

altro3:
@andriy-dmytruk I checked sample and your test - all is ok

PR #1179, opened 2023-08-26
Swagger UI 5.4.2 Redoc 2.1.1

### LINKED ISSUE #1178: Allow Setting of persistAuthorization for Swagger UI Authorization
### Feature description

When integrating Swagger UI through Micronaut OpenAPI, there is currently no straightforward way to set the [`persistAuthorization`](https://github.com/swagger-api/swagger-ui/blob/HEAD/docs/usage/configuration.md) property to allow persisting of authorization data.

If I manually add it to the generated SwaggerUIBundle it works and API auth tokens are being persisted using the browsers `localStorage`.

```js
 ui = SwaggerUIBundle({
                    url: contextPath + '/swagger/demo-api-1.0.yml',
                    persistAuthorization: true,
                    dom_id: '#swagger-ui',
                    presets: [
                        SwaggerUIBundle.presets.apis,
                        SwaggerUIStandalonePreset
                    ],
                    plugins: [
                        SwaggerUIBundle.plugins.DownloadUrl,
                        f
                    ],
                    
                    
                    deepLinking: true,
layout: "StandaloneLayout",
tagsSorter: "alpha",
validatorUrl: null
                });
```

It would be nice if this property could be set directly in the `openapi.properties`

altro3:
Also:

 - Allow null values for extensions properties in openapi object
 - Added Swagger UI property persistAuthorization

Fixed #1178


sdelamo (changes requested):

openapi/src/main/java/io/micronaut/openapi/view/SwaggerUIConfig.java
```diff
@@ -98,6 +98,7 @@ final class SwaggerUIConfig extends AbstractViewConfig {
         VALID_OPTIONS.put("supportedSubmitMethods", AbstractViewConfig::asString);
         VALID_OPTIONS.put("validatorUrl", AbstractViewConfig::asQuotedString);
         VALID_OPTIONS.put("withCredentials", AbstractViewConfig::asBoolean);
+        VALID_OPTIONS.put("persistAuthorization", AbstractViewConfig::asBoolean);
```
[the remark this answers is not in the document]
altro3: done

graemerocher (approved):

openapi/src/main/java/io/micronaut/openapi/swagger/core/jackson/mixin/OpenAPIMixin.java
```diff
@@ -36,6 +37,7 @@
 public abstract class OpenAPIMixin {
 
     @JsonAnyGetter
+    @JsonInclude
```
sdelamo, line 40: is `@JsonInclude` necessary? Do we have any test to verify it is necessary?

altro3: @sdelamo Yeas, it necessary. We don't need to test it. These classes just copied from swagger-core project. This change i got from here: https://github.com/swagger-api/swagger-core/pull/4468

I keep these classes up to date all the time. Why did I copy this piece from swagger-core? For two reasons:

1. This allowed me to remove a bunch of unnecessary dependencies from the project, i.e. from swagger-core, we used only the jackson settings.

2. Now there is a possibility to customize logic, due to this we managed to fix a couple of problems in serialization / deserialization.

sdelamo (approved):

PR #1212, opened 2023-09-17
Remove duplicate elements from final OpenApi object.

### LINKED ISSUE #1204: Duplicated server when generating the openapi file in a custom folder
### Expected Behavior

doing gradle clean build several times should not duplicate the server.

```
servers:
- url: http://localhost:8080
```

### Actual Behaviour

the server is duplicated in the generated api.yml doc;

```
servers:
- url: http://localhost:8080
- url: http://localhost:8080
```

if you keep doing gradle clean build, it will keep adding a new line.

### Steps To Reproduce

1. I added openapi dependencies

```
    annotationProcessor("io.micronaut.openapi:micronaut-openapi:5.0.0-M4")
    compileOnly("io.micronaut.openapi:micronaut-openapi:5.0.0-M4")
    implementation("io.swagger.core.v3:swagger-annotations:2.2.15")
```

2. the server is added to the Application.groovy:

```
@Server(url = 'http://localhost:8080')
class Application {
    static void main(String[] args) {
        Micronaut.run(Application, args)
    }
}
```
3. create a openapi.properties file and change the target file:
`micronaut.openapi.target.file=./docs/openapi/api.yml
`
4. run gradle clean build 2 times or more. The first file will be ok, the second will duplicate the line and so on.

### Environment Information

jdk 17.

### Example Application

_No response_

### Version

4.1.0

altro3:
Fixed #1204

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
@@ -38,6 +38,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.BiConsumer;
+import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java
```diff
@@ -772,12 +773,117 @@ private OpenAPI postProcessOpenApi(OpenAPI openApi, VisitorContext context) {
...
 
         removeEmtpyComponents(openApi);
+        findAndRemoveDuplicates(openApi);
 
         openApi = resolvePropertyPlaceHolders(openApi, context);
 
         return openApi;
     }
 
+    private <T> List<T> findAndRemoveDuplicates(List<T> elements, BiPredicate<T, T> predicate) {
```
graemerocher, line 783: would this not be better written with `elements.stream().distinct()` or `elements.stream().filter(..)`?

altro3: @graemerocher Not for everyone... Personally, I try not to use streams because of their slowness and the creation of a bunch of additional objects. Yes, it is more convenient in terms of syntax, but always worse in performance

graemerocher (approved):

PR #1213, opened 2023-09-17
Add ability to genearate Flux<?> instead Mono<List<?>> for arrays

### LINKED ISSUE #1200: openapi-generator have the option to return a Flux instead of Mono for arrays
### Feature description

Provide the ability to return Flux<Object> instead of Mono<List<Object>>.

When using the openapi-generator to create a server (or client) from an existing spec file, it currently defaults to using "reactive" so will generate everything as Mono's including where it's defined to return a List.  What I'd like is for the ability to return a Flux instead.

For example, the below maven plugin definition:
```
<execution>
  <goals>
    <goal>generate-openapi-server</goal>
  </goals>

  <configuration>
    <enabled>true</enabled>
    <definitionFile>${project.basedir}/server.yaml</definitionFile>
  </configuration>
</execution>
```

Against this spec file:
```
openapi: 3.0.1
info:
  title: Hello
  version: '1.0'

paths:
  /getStuff:
    get:
      operationId: getStuff
      responses:
        default:
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
```

Will generate the code:
```
@Get("/getStuff")
@Produces({"application/json"})
Mono<List<String>> getStuff();
```

What I'd prefer to see if it returning a Flux, as we have some services that return thousands of records which won't fit into memory.
```
@Get("/getStuff")
@Produces({"application/json"})
Flux<String> getStuff();
```

For the spring implementation under [OpenAPITools/openapi-generator](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaSpring/formParams.mustache) they have some logic to check the type isArray to use a Flux over a Mono, but I can't find the equivalent in this repo at the moment.

altro3:
Fixed #1200

Also Fixed bugs in generated tests and removed duplicates model_test templates.


graemerocher:
note that `Flux` is less efficient in general

altro3:
Perhaps, but I don't see any reason why we shouldn't let the user choose.

PR #1217, opened 2023-09-20
Update to 4.1.1 & Rewrite MessageBodyWriter

sdelamo:
I had to rewrite the `DatedResponseBodyWriter` and `PageBodyWriter` when upgrading to Micronaut Framework 4.1.1 

Was the previous usage of `createSpecific` API a correct usage of the API?.

### DIFF
--- a/test-suite-server-generator/src/main/java/io/micronaut/openapi/test/dated/DatedResponseBodyWriter.java
+++ b/test-suite-server-generator/src/main/java/io/micronaut/openapi/test/dated/DatedResponseBodyWriter.java
@@ -1,8 +1,6 @@
 package io.micronaut.openapi.test.dated;
 
-
 import io.micronaut.context.exceptions.ConfigurationException;
-import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.annotation.Order;
 import io.micronaut.core.type.Argument;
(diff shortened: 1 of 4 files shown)


test-suite-server-generator/src/main/java/io/micronaut/openapi/test/dated/DatedResponseBodyWriter.java
```diff
@@ -66,27 +40,23 @@ public void writeTo(
...
     ) throws CodecException {
-        if (bodyType != null && bodyWriter != null) {
-            if (dated.getLastModified() != null) {
-                headers.add(LAST_MODIFIED_HEADER, dated.getLastModified().toString());
-            }
-            bodyWriter.writeTo(bodyType, mediaType, dated.getBody(), headers, outputStream);
-        } else {
-            throw new ConfigurationException("No JSON message writer present");
+        Argument<T> bt = type.getTypeParameters()[0];
+        MessageBodyWriter<T> writer = registry.findWriter(bt, List.of(MediaType.APPLICATION_JSON_TYPE))
```
yawkat, line 44: findWriter should be cached using createSpecific instead of doing it inside writeTo.

PR #1233, opened 2023-10-01
New module: openapi-adoc

altro3:
Introduce new module to convert openAPI json and yml format to Asciidoc format. After it you can convert it to Markdown, HTML, PDF or DOCX.

Base template looks like this (PDF format):

изображение

изображение

**You can customize full template or only part of it.**

I integrated it with annotation processor as a optional module. To use it you just need add dependency to annotation processor classpath:
```groovy
annotationProcessor("io.micronaut.openapi:micronaut-openapi-adoc")
```

**Need to merge it first:** https://github.com/micronaut-projects/micronaut-build/pull/623


altro3:
@graemerocher and what about this new module? Created 2 weeks ago, but nobody review it :-(

openapi-adoc/src/main/java/io/micronaut/openapi/adoc/utils/CollectionUtils.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * CollectionUtils.
+ *
+ * @since 5.2.0
+ */
+public final class CollectionUtils {
```
graemerocher, line 26: is this needed there is already this type in `micronaut-core` module

openapi-adoc/src/main/java/io/micronaut/openapi/adoc/utils/FileUtils.java
```diff
@@ -0,0 +1,60 @@
...
+public final class FileUtils {
+
+    public static final String FILE_SCHEME = "file:";
+    public static final String CLASSPATH_SCHEME = "classpath:";
+
+    private FileUtils() {
+    }
+
+    public static String loadFileFromClasspath(String location) {
+        String file = location.replaceAll("\\\\", "/");
```
graemerocher, line 35: need to check there are no path traversal vulnerabilities here. //cc @yawkat

openapi-adoc/src/main/java/io/micronaut/openapi/adoc/utils/SwaggerUtils.java
```diff
@@ -0,0 +1,82 @@
...
+            return mapper.readValue(swaggerFileContent, OpenAPI.class);
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException("Can't parse swagger file", e);
+        }
+    }
+
+    public static OpenAPI readOpenApiFromLocation(String location) {
+
+        var isJson = location.endsWith(".json");
+        var adjustedLocation = location.replaceAll("\\\\", "/").toLowerCase();
```
graemerocher, line 63: Same thing here with path traversal

graemerocher (changes requested):

graemerocher:
See https://github.com/micronaut-projects/micronaut-object-storage/commit/ac16eb7502d4d3120d82f6a021f8eb73576c4631#diff-20cca42a45e0a0ce9f0ca51935262795061b156de0b3b5940c7b9f89517583c7R227 for how to prevent path traversal

graemerocher:
Sorry for the delay in reviewing PRs, appreciate all your work which is a significant contribution!

openapi-adoc/src/main/java/io/micronaut/openapi/adoc/utils/CollectionUtils.java
altro3: Yes, I know, but the idea is to not depend on this library.

By the way, I saw that it will still appear in the dependencies, because... it is added by the Build plugin. How to correctly remove this background dependency?

same problem in other modules: `openapi-annotations`, `openapi-generator`...

openapi-adoc/src/main/java/io/micronaut/openapi/adoc/utils/SwaggerUtils.java
altro3: Hm.. i still don't understand, what problem with my code? Your example is not the same, as my case. My location can be in URI format, can be just file system path, can be pah inside jar or classpath. This is the easiest and fastest way to get rid of problems with different slashes in the path and OS

graemerocher: ok since this is build time only it is probably fine

graemerocher (approved):

PR #1243, opened 2023-10-07
Fix setting generateHttpResponseWhereRequired option

infomouse:
Setting generateHttpResponseWhereRequired option does not work because it is not passed to codeGenerator.


altro3:
@infomouse Check and fix tests, pls

infomouse:
> @infomouse Check and fix tests, pls
Fixed.

PR #1268, opened 2023-10-28
fix group properties not being read from openapi.properties

### LINKED ISSUE #1267: Group properties not being read
### Expected Behavior

Group properties set in `openapi.properties` should be read properly and respected.

### Actual Behaviour

Group properties are not read properly. 

### Steps To Reproduce

- Add group properties to a `openapi.properties` file as defined here: https://micronaut-projects.github.io/micronaut-openapi/5.1.0/guide/index.html#groups.
- Add an @OpenAPIGroup annotation to an endpoint controller as defined in the docs linked above.
- OpenAPI/Swagger files generated do not respect defined groups.

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.9.2

danielvinihan:
Group properties such as `micronaut.openapi.groups.*.packages=***` are not being read from the `openapi.properties` file, and therefore the properties are not being respected. The `readGroupsProperties` function passes the property name, to `setGroupProperty` but it is currently passing the full name, i.e. `micronaut.openapi.groups.*.packages`, when it should just be passing the key after the group name, i.e. `packages`. 

Fixes #1267


altro3:
@graemerocher Could you merge this fix?

graemerocher (approved):

PR #1274, opened 2023-11-03
Add `visible = true` to `@JsonTypeInfo` for micronaut-serde

### LINKED ISSUE #1273: OpenAPI Generator always generates the `@JsonTypeInfo` annotation without the `visible` property in models for Micronaut Serialization
### Expected Behavior

**_IMPORTANT!!!_** The fix for this issue should not be released, before the fix of Micronaut Serialization to support the `visible` property for the `@JsonTypeInfo` annotation. This was fixed today (3rd Nov 2023) in [this issue](https://github.com/micronaut-projects/micronaut-serialization/issues/581). 

I expect the OpenAPI Generator to generate the `@JsonTypeInfo` annotation so it is identical for Jackson and Micronaut Serialization. So if a `visible` property is generated for Jackson, it should also be generated for Micronaut Serialization.

### Actual Behaviour

The `visible` property is never generated for the `@JsonTypeInfo` annotation in models for Micronaut Serialization.

### Steps To Reproduce

1. Clone the repository https://github.com/runenielsen/mn-discriminator-null-serde-bug, that uses Micronaut Serialization.
2. Run the `testPostModelWithDiscriminator` test case in `BookInfoControllerTest` to post the model with a discriminator to a service endpoint with validation.
3. See that the test fails due to a validation failure "bookInfo.type: must not be null".
4. Inspect the generated `BookInfo` model in `generateServerOpenApiModels`and see that the `@JsonTypeInfo` annotation looks like this: `@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")`
5. To run the test with the correct model, comment in the code under `/* Comment this in to run with the visible property set to true as in the generated models for Jackson.` in `build.gradle.kts`. Notice this currently gives this error due to `visible` not being supported in Micronaut Serialization: `/Users/rune.nielsen/projects/mn-discriminator-null-serde-bug/build/generated/openapi/generateServerOpenApiModels/src/main/java/com/example/openapi/server/model/BookInfo.java:50: error: Annotation @JsonTypeInfo specifies attribute 'visible'. Currently supported attributes include: [include, use, property, defaultImpl] public class BookInfo {`. This is why the fix here should not be released before the fix already done in [this issue](https://github.com/micronaut-projects/micronaut-serialization/issues/581).
7. Now clone the repository https://github.com/runenielsen/mn-discriminator-not-null-with-jackson, that contains the same code, but uses Jackson Databind instead of Micronaut Serialization.
8. See that the same test succeeds in that repository.
9. Inspect the generated `BookInfo` model in `generateServerOpenApiModels`and see that the `@JsonTypeInfo` annotation looks like this: `@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type", visible = true)`

### Environment Information

- macOS
- JDK 17 (Temurin)

### Example Applications

https://github.com/runenielsen/mn-discriminator-null-serde-bug and https://github.com/runenielsen/mn-discriminator-not-null-with-jackson

### Version

4.1.6

altro3:
Fixed #1273


runenielsen (approved):

dstepanov:
Actually, this cannot be always true. If the subtype model doesn’t expect it or doesn’t define ignore unknown it will fail

altro3:
@dstepanov @graemerocher Not about this PR. So, I tried to publish micronaut-openapi 5.2.0, but publishing failed. Could you tell me what is wrong? https://github.com/micronaut-projects/micronaut-openapi/actions/runs/6745652423

dstepanov:
The discriminator field doesn't need to be in the class.
The idea of making it visible is for the deserializer to consume it. If the deserializer is not expecting it, you can get an exception.
It should be added only for this kind of case https://github.com/runenielsen/mn-discriminator-not-null-with-jackson/blob/main/src/main/resources/modelwithdiscriminator.yml Discriminator property name `propertyName` `type` is also included in the subtype.

dstepanov:
Looks like the release is published but something failed.

altro3:
@dstepanov maybe, some troubles with new module micronaut-openapi-adoc ?

dstepanov:
Please make sure the internal build plugin is the latest one

altro3:
@dstepanov @graemerocher please, review and merge it, after it I can finish issue with ksp

dstepanov (approved):

PR #1298, opened 2023-11-09
Fix generated method toString with inheritance.

altro3:
Add no args constructor for POJOs, when use jackson-databind.


altro3:
About toString: as I seee, this i bug in openapi-generator lib. `allVars` property must contains all properties for class plus all parent properties, but now in some cases it contains only parent properties.

About no-args constructor: need it for jackson-databind, because jackson by default use no-args constructor or trying to find method with `@JsonCreator` annotation. This is easiest solution. The second solution: add annotation `@JsonCreator` to required args constructor, but in this case you also need to add `@JsonProperty` annotation for every constructor argument, because jackson can't automatticaly map json proeprties to method arguments

graemerocher (approved):

altro3:
@andriy-dmytruk Hi! Could you revie wit? And after this PR we can finalize next release

openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
```diff
@@ -382,9 +387,9 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#vendorE
     @Override
     public String toString() {
         return "{{classname}}("
-        {{#allVars}}
+        {{#vendorExtensions.allVars}}
```
andriy-dmytruk, line 390: I think we should fix the `allVars` property instead of creating a new property in vendorExtensions.

openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
```diff
@@ -131,6 +131,11 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#vendorE
     {{/vars}}
 
     {{#requiredPropertiesInConstructor}}
+        {{^micronaut_serde_jackson}}
+    public {{classname}}() {
```
andriy-dmytruk, line 135: I think it would be better for constructor to be private or package-private. Otherwise the point of required properties is lost.

andriy-dmytruk (changes requested):

openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
altro3: fixed

openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache
altro3: fixed

altro3:
Also removed `@Introspected` annotation, when we added `@Serdeable` annotation

andriy-dmytruk (approved):

PR #1328, opened 2023-11-30
Fix map gradle plugin properties to generator properties.

### LINKED ISSUE #1325: MicronautCodeGeneratorEntryPoint clientOptions are not propagated to DefaultGenerator via JavaMicronautClientCodegen
### Expected Behavior

When passing for example additionalClientTypeAnnotations in ClientOptions to MicronautCodeGeneratorEntryPoint, the annotations are generated in the corresponding API interface.

### Actual Behaviour

When passing for example additionalClientTypeAnnotations in ClientOptions to MicronautCodeGeneratorEntryPoint, the annotations are NOT generated in the corresponding API interface.
This behaviour is true for various client options as well as other general and server options.

This is because the OpenAPI DefaultGenerator only tries to get these information from additionalProperties.

For example MicronautCodeGeneratorEntryPoint sets additionalClientTypeAnnotations in JavaMicronautClientCodegen's additionalClientTypeAnnotations NOT in additionalProperties.
The processOpts() method(s) never set additionalClientTypeAnnotations, clientId etc. to additionalProperties.
So these client configs are never passed to OpenAPI DefaultGenerator and as a result they never get generated in the API interface.

I've extended the OpenApiClientGeneratorSpec in micronaut-openapi-plugin and debugged all the way to the DefaultGenerator class.
The properties handling in MicronautCodeGeneratorEntryPoint and JavaMicronautClientCodegen doesn't seem consistent with what DefaultGenerator expects. A couple of properties one can configure for example in micronaut-openapi-plugin's OpenApiSpec, OpenApiClientSpec, OpenApiServerSpec don't get propagated to DefaultGenerator and therefore are not generated.
The tests only tests for TaskOutcome.SUCCESS which isn't sufficient. 

### Steps To Reproduce

Configure a micronaut project with
micronaut {
                openapi {
                    client(file("petstore.json")) {
                        additionalClientTypeAnnotations = [ '@Header(name="Api-Key", value="somekey")' ]
                        clientId = "someid"
                    }
                }
            }
and generate the code.

### Environment Information

All OS, all JDKs

### Example Application

_No response_

### Version

4.2.0

altro3:
Fixed #1325


altro3:
@sdelamo fixed.

altro3:
@andriy-dmytruk @melix This is critical bug, becuase now users can set some configuration properties from gradle plugin. need to reveiew it, merge and release it before new gradle plugin version will be released

melix (changes requested):
Stringly based APIs are really awful. I don't feel like this is the right place to fix this. The builders should _not_ use `writePropertyBack`. Instead this should be done in `processOpts`. The idea of the builders was to get rid of these nasty string maps at some point.

altro3:
@melix Is everything right now?

melix (approved):

melix:
Yes, thank you! We're trying to keep our builders free of OpenAPI generator's mess, it's much cleaner now :+1:

melix:
Looks like there are failing tests again.

andriy-dmytruk (approved):

altro3:
@sdelamo Now all work is completed. I think, we can merge all PRs and finilize release

PR #1329, opened 2023-11-30
extract method to avoid duplication

sdelamo:


altro3 (approved):

openapi/src/main/java/io/micronaut/openapi/view/RapiPDFConfig.java
```diff
@@ -91,6 +91,11 @@ public String toString() {
 
     boolean enabled; //false
 
+    @Override
+    public boolean isEnabled() {
+        return enabled;
+    }
+
```
altro3, line 98: @sdelamo Wrong order: methods must be after constructors

PR #1380, opened 2024-01-13
Remove micronaut-http and micronaut-inject from dependencies

altro3:


graemerocher:
What is the motivation for this?

altro3:
I just want to minimize dependencies. There is also an idea to add support for Spring annotations. I don't have an exact solution yet, but I decided to first get away from the tight connection with micronaut-http and switch specifically to string class names. The next step is to embed spring support. Perhaps I will put the class names in different interfaces as lists of constants, maybe I will write a separate visitor. I haven't worked out a solution yet.

In general, it will be easier if only the necessary things really remain from mikronaft - that is, the core.

graemerocher:
I don't think we should be duplicating annotations. Spring support can already be achieved by mapping the spring annotations to Micronaut versions

altro3:
Maybe. I haven't checked yet.

In any case, I don't see the point in leaving these two libraries in dependencies. because they are not required in the code. I only need the kernel for the library to work correctly. Fewer dependencies - fewer problems

graemerocher:
I disagree. The PR duplicates many annotations and big classes like UriTemplate and creates a code maintenance problem. I am afraid I can't approve this PR.

PR #1382, opened 2024-01-14
Fix lost enum classes when it unnamed schemas for endpoint parameters

altro3:


graemerocher (approved):

altro3:
The error of the openapi generator has also been fixed: 'path.parameters` were lost - in this case, these parameters should be included in each path operation

PR #1416, opened 2024-02-04
Add ksp flag to kotlin generator to generate only KSP compatible classes

altro3:
It means without prefix `@field:` for annotaions on constructor parameters level


graemerocher (approved):

altro3:
@dstepanov merge it, pls

PR #1427, opened 2024-02-09
Add support openapi-generator mapping configuration properties:

### LINKED ISSUE #1421: OpenAPI Generator: Expose configuration for import mapping / schema mapping
### Feature description

Hi,

OpenAPITools support customization by schema mapping and import mapping:

See: https://openapi-generator.tech/docs/customization/#schema-mapping
and: https://openapi-generator.tech/docs/customization/#bringing-your-own-models

It seems that these two options are already used internally, but are not made available for external configuration (especially for the gradle plugin).

It would be very helpful to provide these two options to reuse models from your own libraries instead of having them generated. For example, you can use existing models for error responses.

altro3:
schemaMapping
importMapping
nameMapping
typeMapping
enumNameMapping
modelNameMapping

Fixed #1421


sdelamo (commented):
can you include tests?

altro3:
@sdelamo Well, what can I say, if I write a regular unit test, then it will not be testing our generator, but the openapi-generator library, because it performs all these functions. I'm not sure it's worth spending time on this.

I just added the ability to set these parameters through the configuration so that the user can set them through the plugin.

So, in order to fully test this mapping, you need to put all these properties in the gradle, but because of this, the generated code will completely change - all class names, mapping to other schemes, etc. If you add these tinctures to existing test suites, you will have to completely recycle them, because absolutely all names and the main the problem is, the generated code will simply be difficult to match with the source file.

I checked all this locally, everything worked out correctly, added only one property to the test - 
```groovy
nameMapping = [test: "changedTest"]
```
because it affects only one class. Prefixes and suffixes affect absolutely all generated code

openapi-generator/src/main/java/io/micronaut/openapi/generator/MicronautCodeGeneratorEntryPoint.java
```diff
@@ -141,6 +142,46 @@ private void configureOptions() {
...
+            if (options.inlineSchemaNameMapping != null) {
+                javaCodeGen.addInlineSchemaNameMapping(options.inlineSchemaNameMapping);
+            }
+            if (options.inlineSchemaOption != null) {
+                javaCodeGen.addInlineSchemaOption(options.inlineSchemaOption);
+            }
+            if (options.openapiNormalizer != null) {
+                javaCodeGen.addOpenapiNormalizer(options.openapiNormalizer);
+            }
+            if (options.apiNamePrefix != null && !options.apiNamePrefix.isBlank()) {
```
sdelamo, line 172: We have a utils class `StringUtils::isNotBlank`

sdelamo (approved):

openapi-generator/src/main/java/io/micronaut/openapi/generator/MicronautCodeGeneratorEntryPoint.java
altro3: I know, but this module does not depend on the libraries of the micronaut. It depends only on the openapi-generator

PR #1433, opened 2024-02-10
Add @Retention(RUNTIME) annotation to HardNullable

averak:
Currently, the `@HardNullable` annotation is missing the `@Retention(RUNTIME)` specification.  By adding `@Retention(RUNTIME)`, we unlock the full potential of the `@HardNullable` annotation, allowing it to be utilized more effectively in Micronaut applications and libraries that depend on runtime annotation processing.

Thank you to all the maintainers for your dedication and hard work that make contributions like mine possible.


sdelamo:
I created a PR to deprecate it https://github.com/micronaut-projects/micronaut-openapi/pull/1434

Thanks for the contribution @averak However, if we move to use `@Runtime(inherited=true)` instead of `@HardNullable` it would make sense to close this PR.

averak:
@graemerocher @sdelamo 

Thank you for your reaction!!
I am looking forward to being able to retrieve @Nullable annotation information at runtime.

PR #1445, opened 2024-02-18
Fix uriMatchTemplates for mapping annotation processors (like micronaut-jaxrs-processor)

altro3:
See https://github.com/micronaut-projects/micronaut-core/issues/4983


sdelamo (changes requested):
Can you add tests?

altro3:
@sdelamo Here is the test app: https://github.com/micronaut-projects/micronaut-core/issues/4983.

But now I see that the problem may have already been solved. But I would leave this fix just in case. As I said, I encountered a similar problem when using micronut-spring - for some reason the headers are mapped to the HttpHeaders annotation, although they should be in HttpHeader.

I had to add a code that takes this situation into account. I would do the same with the Controller annotations, as suggested by the author of the issue

sdelamo:
I don't think this PR is necessary. I create a [test which reproduces it behaves correctly for JAX-RS](https://github.com/micronaut-projects/micronaut-openapi/pull/1446).

sdelamo:
Please reopen if other issues need this fix. Those issues should be reproducible with a test.

PR #1459, opened 2024-03-01
Fix mapping entry point generator properties to client properties

altro3:
See https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/949


sdelamo (changes requested):
Please, add a test

altro3:
@sdelamo done

sdelamo (approved):

PR #1467, opened 2024-03-06
Fixes for generating openAPI clients

altro3:
- Add ability to disable generating `path` attribute in Client annotation
 - Changed basePathSeparator default value to `.` symbol
 - Add suffixes by clientId to client base path placeholders

With `clientId` it will be:
```java
@Client("myclient")
```
But you could set `clientPath = true` and you will see this:
```java
@Client(id = "myclient", path = "${myclient.base-path}")
```
Without client id it will be
```java
@Client("${openapi-micronaut-client.base-path}")
````


altro3:
See https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/949#issuecomment-1978619863

andriy-dmytruk (approved):

PR #1491, opened 2024-03-21
Update HelloController.kt

adrianofsp:
fix missing annotation parameter name identifier.

### DIFF
--- a/docs-examples/example-kotlin/src/test/kotlin/io/micronaut/configuration/openapi/docs/controllers/annotations/HelloController.kt
+++ b/docs-examples/example-kotlin/src/test/kotlin/io/micronaut/configuration/openapi/docs/controllers/annotations/HelloController.kt
@@ -25,11 +25,11 @@ open class HelloController {
     @Operation(summary = "Greets a person", description = "A friendly greeting is returned")
     // Please Note: Repeatable Annotations with non-SOURCE retentions are not yet supported with Kotlin, so we are using `@ApiResponses`
     // instead of `@ApiResponse`, see https://youtrack.jetbrains.com/issue/KT-12794
-    @ApiResponses(
+    @ApiResponses(value = [
             ApiResponse(content = [Content(mediaType = "text/plain", schema = Schema(type = "string"))]),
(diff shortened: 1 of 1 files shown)


altro3:
With KSP it is work fine. It's strange why the library works correctly and the compiler doesn't complain....
изображение
изображение

But I think this solution will be better:

BEFORE:
```kotlin
    @Get(uri = "/greetings/{name}", produces = [MediaType.TEXT_PLAIN])
    @Operation(summary = "Greets a person", description = "A friendly greeting is returned")
    // Please Note: Repeatable Annotations with non-SOURCE retentions are not yet supported with Kotlin, so we are using `@ApiResponses`
    // instead of `@ApiResponse`, see https://youtrack.jetbrains.com/issue/KT-12794
    @ApiResponses(
            ApiResponse(content = [Content(mediaType = "text/plain", schema = Schema(type = "string"))]),
            ApiResponse(responseCode = "400", description = "Invalid Name Supplied"),
            ApiResponse(responseCode = "404", description = "Person not found")
    )
    @Tag(name = "greeting")
    open fun greetings(name: String): Mono<String> {
        return Mono.just("Hello $name, how are you doing?")
    }
```

AFTER:

```kotlin
    @Tag(name = "greeting")
    @Operation(summary = "Greets a person", description = "A friendly greeting is returned")
    @ApiResponse(content = [Content(mediaType = "text/plain", schema = Schema(type = "string"))])
    @ApiResponse(responseCode = "400", description = "Invalid Name Supplied")
    @ApiResponse(responseCode = "404", description = "Person not found")
    @Get(uri = "/greetings/{name}", produces = [MediaType.TEXT_PLAIN])
    open fun greetings(name: String): Mono<String> {
        return Mono.just("Hello $name, how are you doing?")
    }
```

Need to remove comment about `https://youtrack.jetbrains.com/issue/KT-12794` - it's already fixed

@adrianofsp WDYT?

adrianofsp:
Hi,

It’s true the compiler does not complain and it works.
I forgot to include one dependency on the build.

Thanks

altro3:
Ok, I fixed this example here https://github.com/micronaut-projects/micronaut-openapi/pull/1492
изображение

PR #1552, opened 2024-04-24
Fix inline resolver for open api generator

altro3:
This is more corect solution for openapi generator problem. This is fix for bug in openap genertor: problem with parameters, when it collection of enum - openapi generator lost this enum and don't generate it. Yesterday I fixed it with internal enums inside client interfaces and controller interfaces, this solution works, but it's not entirely correct. It is necessary to create separate files for such enums so that this logic does not differ from the rest. Today I fixed` openapi-generator/src/main/java/io/micronaut/openapi/generator/MicronautInlineModelResolver.java` - Now it handles this case correctly.

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
@@ -691,6 +691,11 @@ public abstract class AbstractMicronautJavaCodegen<T extends GeneratorOptionsBui
     }
 
     @Override
+    public String toApiName(String name) {
+        return Utils.toApiName(name, apiNamePrefix, apiNameSuffix);
+    }
+
+    @Override
     public void addOperationToGroup(String tag, String resourcePath, Operation operation, CodegenOperation co,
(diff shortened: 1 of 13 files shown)


openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
```diff
@@ -37,7 +37,7 @@ protected String generateFiles(MicronautCodeGenerator<?> codegen, String configP
         } catch (IOException e) {
             fail("Unable to create temporary directory for output");
         }
-        output.deleteOnExit();
+//        output.deleteOnExit();
```
sdelamo: is this commented on purpose.

sdelamo (changes requested):
Can you add a PR description about what this change is about?

openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
altro3: fixed

altro3:
> Can you add a PR description about what this change is about?

done

sdelamo (approved):

PR #1610, opened 2024-06-30
Add support extra Schemas

### LINKED ISSUE #1598: Allow the ability to register extra models to the OpenAPI schema
### Feature description

Currently only models used by endpoints get exported into the OpenAPI schema. There are use cases (such as using the [OpenAPI code generator](https://github.com/OpenAPITools/openapi-generator) to generate models for downstream consumption) where registering extra models is necessary. 

It would be great if it was possible to get a handle on the OpenAPI schema object and arbitrary models could be registered.

altro3:
Move schema definition logic to different class

Fixed #1598


openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * A {@link TypeElementVisitor} that builds the extra Open API schema definitions included by @{@link OpenAPIExtraSchema}
+ * at the compile time.
+ *
+ * @since 6.12.0
+ */
+@Internal
+@SupportedOptions(MICRONAUT_OPENAPI_ENABLED)
+public class OpenApiExtraSchemaVisitor implements TypeElementVisitor<Object, Object> {
```
graemerocher: override `getSupportedAnnotationNames` and provide the annotation names this processes

graemerocher: override `getVisitorKind` and return `ISOLATING` since this processes just a single annotation

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+            if (isNotEmpty(excludeClasses)) {
+                excludedExtraSchemaClassNames.addAll(Arrays.asList(excludeClasses));
+            }
+            if (isNotEmpty(excludePackages)) {
+                excludedExtraSchemaPackages.addAll(Arrays.asList(excludePackages));
+            }
+            if (isNotEmpty(packages)) {
+                for (var packageName : packages) {
+                    var classEls = findAllClassesUsingClassLoader(packageName, context);
```
graemerocher: you should never use class loaders this way in an annotation processor. You should call `visitorContext.getClassElements(..)` to search for classes

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
```diff
@@ -0,0 +1,207 @@
...
+ * A {@link TypeElementVisitor} that builds the extra Open API schema definitions included by @{@link OpenAPIExtraSchema}
+ * at the compile time.
+ *
+ * @since 6.12.0
+ */
+@Internal
+@SupportedOptions(MICRONAUT_OPENAPI_ENABLED)
+public class OpenApiExtraSchemaVisitor implements TypeElementVisitor<Object, Object> {
+
+    private static Map<String, Schema> extraSchemas = new HashMap<>();
```
graemerocher: use `LinkedHashMap` otherwise the output will change order each time you compile breaking compilation reproducibility

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
graemerocher: the first generic argument should also be `OpenApiExtraSchema`

altro3: if I set `ISOLATING` kind, it doesn't work

altro3: The same situation as with VisitorKind - doesn't work, if I use `TypeElementVisitor<OpenAPIExtraSchema, Object>`

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
altro3: Then it turns out that now there is a problem. There is no way to get a list of all classes in a package.

I tried using the method you suggested, but it doesn't work. The only way I found to get a list of all classes is to use classLoader

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
altro3: fixed

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
graemerocher: it is the way every other visitor that supports scanning packages works (see https://github.com/micronaut-projects/micronaut-core/blob/4.5.x/core-processor/src/main/java/io/micronaut/context/visitor/BeanImportVisitor.java), in fact the way you have written this simply will not work in a real arrangement where you have a separate annotation processor classpath vs compilation classpath.

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
graemerocher: probably because it needs to be the repeatable wrapper `TypeElementVisitor<OpenAPIExtraSchemas, Object>`

altro3: it's true. Fixed

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
altro3: @graemerocher 

The problem is that this method doesn't work. Here's an example:

1. I enter the package `"io.micronaut.openapi.extra"`. This package contains 3 files, and 4 classes (1 internal)

2. The `context.getClassElements(packageName, EMPTY_STRING_ARRAY)` method returns an empty list.

3. `ClassLoader.getSystemClassLoader().getResourceAsStream(packageName.replaceAll("[.]", "/"))` - returns a list of classes

изображение

The second important problem is that there is no wildcard support. When I want to additionally scan everything under packages.

So what should we do? Create a new method in the core module?

graemerocher: it needs to be `context.getClassElements(packageName, "*")`

graemerocher: be aware I am not trying to be difficult here, the way you have implemented this may work in your tests but it will not work in a real project.

altro3: Yes, I understand. I tried using this method, just as I showed, I couldn't get the desired result.

So, I checked the method you suggested (with the * argument). Yes, indeed, everything worked.

BUT! Now all that remains is to solve the wildcard problem - how can I get all possible class elements from a specified package, INCLUDING all sub-packages? As I understand it, this functionality is currently missing

graemerocher: that will require a PR to core I imagine to extend this method. This could be done in a separate PR in the meantime you could just document all package names have to be specified.

altro3: ok,  done

graemerocher:
could you target the 6.12.x branch?

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
```diff
@@ -0,0 +1,215 @@
...
+    public static Collection<String> getExcludedExtraSchemas() {
+        return excludedExtraSchemas;
+    }
+
+    public static void clean() {
+        extraSchemas = new HashMap<>();
+        extraSchemaClassnamesToNames = new HashMap<>();
+        excludedExtraSchemaClassNames = new HashSet<>();
+        excludedExtraSchemas = new HashSet<>();
+        excludedExtraSchemaPackages = new HashSet<>();
```
graemerocher: these also have to be changed to `LinkedHashMap`

graemerocher (changes requested):

openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiExtraSchemaVisitor.java
altro3: fixed

altro3:
> could you target the 6.12.x branch?

@graemerocher done

graemerocher (approved):

graemerocher:
Thanks for the contribution

PR #1647, opened 2024-07-26
Add ability to customize separators for generics and inner classes in schema names

### LINKED ISSUE #1646: Using OpenAPI annotations to create Swagger YAML for generics is not compatible with AWS
### Feature description

From the AWS specs: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html#api-gateway-known-issues-rest-apis

`Model names can only contain alphanumeric characters.`

```java
    @Get("/pageablelist")
    @Operation(responses = {@ApiResponse(responseCode = "200", useReturnTypeSchema = true)})
    @OffsetBasedPageableParameters
    public MyPageableListPage getMyPageableList(@Parameter (hidden = true) OffsetBasedPageable pageable) { /*...*/}
```

Which means, the generated schema:

```
# swagger.yml
components:
  schemas:
    MyPageableListPage:
      $ref: '#/components/schemas/OffsetBasedPage_SomeClazz_'
    OffsetBasedPage_SomeClazz_:
      ...
```
is not valid for AWS.

Which results in:
```console
Value 'Errors found during import: Unable to create model for 'OffsetBasedPage_aaa_': Model name must be alphanumeric: OffsetBasedPage_aaa_
Unable to insert model 'aaaPage' due to an unsatisfied schema reference.
```

I tried to change the schemas naming via:
https://micronaut-projects.github.io/micronaut-openapi/latest/guide/#schemasNaming

```
                        <arg>-J-Dmicronaut.openapi.property.naming.strategy=UPPER_CAMEL_CASE</arg>
```
but this only affects the `properties` and not the `schemas`.

## Proposal:

So I propose to add the option to change the naming strategy for schemas, too:
```
micronaut.openapi.schema.naming.strategy=UPPER_CAMEL_CASE
```

**Reference:**
https://stackoverflow.com/questions/37572531/amazon-api-gateway-import-from-swagger-error-not-taking-generics

altro3:
Fixed #1646


Gustl22:
Very nice! That was fast! Just wondering about the naming descision. There already exists this:
```
micronaut.openapi.property.naming.strategy
```

Wouldn't make it more sense to name it like 
```
micronaut.openapi.schema.name.separator.[...]*
```
This also leaves more possibilities to change other attributes of the `schema` than just the name.

altro3:
The property naming strategy is Jackson settings. By default, properties in schemas are named the same as properties in classes, but a different property naming strategy is often used when serializing objects. And this directly affects the API format

The name of the scheme is a convention. The name of the scheme does not in any way affect the communication protocol between the client and the server; this is simply necessary to improve the readability of openapi specification. By default, the name scheme matches the class names, the only characters we use other than numbers and letters are `_` and `.` .  The settings I added are sufficient to bring the schema names to the `"[a-zA-Z0-9]+"` pattern.

In addition, you can set custom schema names using swagger annotations.

What I mean is that there are no strategies for naming schemes; they always coincide with class names.

You are talking about some settings of schemas. Ok, be more specific with your proposal, at the moment I can’t imagine what else might be needed.

graemerocher (approved):

Gustl22:
@altro3 I think you misunderstood me, the implementation and the functionality is perfectly fine. I just don't understand, why the new setting in this PR is called `micronaut.openapi.schema-name.separator.empty` instead of `micronaut.openapi.schema.name.separator.empty`. As that may also is in the same scope of `schema`, something like `micronaut.openapi.schema.decorators`.

altro3:
@Gustl22 Oh... got it, renamed it to `micronaut.openapi.schema.name.*`

altro3:
@graemerocher could you set branch 6.12.x as default?

PR #1677, opened 2024-08-06
Fix generator for controller parameters with arrays and enums

altro3:


altro3:
This is foolproof , when in `array schema` also added `enum` property

sdelamo (approved):

PR #1683, opened 2024-08-08
Unwrap allOf block when it possible.

### LINKED ISSUE #1680: `@Schema` annotation on DTO field generates an OpenAPI spec with description (readOnly, requiredMode, etc) under allOf instead of on the property
### Expected Behavior

Annotating a field with `@Schema(description = "abc")` should result in the description in the generated OpenApi spec being on the component model property, the same as if the field had a JavaDoc comment.

### Actual Behaviour

The generated OpenApi spec puts the description on under `allOf`, where it can't be used by generators as the property description. Also some generators create a subclass just to hold the description, which is not at all the intended effect.

### Steps To Reproduce

```java
@Schema(description = "A simple DTO")
public class MyDto {
  @Schema(description = "A string field")
  private String field1;
  // ...
}

@Schema(description = "A DTO containing the other DTO")
public class MyDto2 {
  @Schema(description = "A field containing another DTO") // <== This description should be on the property not the nested schema
  private MyDto field2;
  // ...
}
```

generates an openapi spec with 
```yaml
components:
  schemas:
    MyDto:
      type: object
      properties:
        field1:
          type: string
          description: A string field
      description: A simple DTO
    MyDto2 :
      type: object
      properties:
        field2:
          allOf:
            - $ref: "#/components/schemas/MyDto"
            - description: A field containing another DTO
      description: A DTO containing the other DTO
```
But it should be:
```yaml
components:
  schemas:
    MyDto:
      type: object
      properties:
        field1:
          type: string
          description: A string field
      description: A simple DTO
    MyDto2 :
      type: object
      properties:
        field2:
          allOf:
            - $ref: "#/components/schemas/MyDto"
          description: A field containing another DTO
      description: A DTO containing the other DTO
```

Note that in the expected output, the description `"A field containing another DTO"` is on the property. In the first one, it's on the allOf schema.

A partial workaround is to remove the `@Schema` annotation and put the description in the JavaDoc for the field. That does generate the correct OpenApi spec in some cases but not in others.

Also, this may not be limited to just `description`. Other schema properties like `required` or `requiredMode` should also be placed on the property, not pushed under `allOf`. For example: `@Schema(requiredMode = RequiredMode.REQUIRED)`.

### Environment Information

Java 21

### Example Application

_No response_

### Version

4.4.3

altro3:
Fixed #1680

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiNormalizeUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiNormalizeUtils.java
@@ -15,14 +15,7 @@
  */
 package io.micronaut.openapi.visitor;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
-import java.util.function.BiConsumer;
-import java.util.function.Function;
-
+import com.fasterxml.jackson.core.JsonProcessingException;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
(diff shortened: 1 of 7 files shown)


graemerocher (approved):

kevin-wise (approved):
Tested with a local build, it seems to work as intended

kevin-wise:
Any idea when this might get merged and released? We would like to upgrade the version of the generator we use, but that depends on this fix.

altro3:
I can't do it myselfl-  haven't rights to merge PRs

@graemerocher ping

kevin-wise:
we will also need https://github.com/micronaut-projects/micronaut-openapi/pull/1687

PR #1687, opened 2024-08-09
Fix parsing UrlTemplates

altro3:


kevin-wise (approved):
Test with a local build, and this does fix the problem, and produced no unexpected changes. Thanks!
Path shows up as `/api/v2/bundles/{bundleId}:refresh` again in the openapi output file.

kevin-wise:
Any idea when this might get merged and released?

PR #1701, opened 2024-08-17
Added schema duplication resolution mode

### LINKED ISSUE #1698: Generated spec contains only one Schema if multiple Schema.name="xy" exist
### Feature description

Consider this situation

```kotlin
@Schema(name="Car")
data class Car1(val p1: String)

@Schema(name="Car")
data class Car2(val p2: String)
```
This will today generate a single Car schema in the openapi spec (while it seems not deterministic which of both types will be used). So one type is missing in the spec. The generated spec is valid and the generator process finishes with status ok, without any hint that some Schemas were overridden.

A similar issue was solved with #947 . But that fix only considers the name of the type and not the annotated Schema.name override.

It would be good to either:
- Generate underscore suffix (as done in #947) based on the Schema.name
- Have a configuration property to make the process error if overrides would happen

altro3:
Fixed #1698

Deprecated configuration properties:

`micronaut.openapi.schema-prefix`
`micronaut.openapi.schema-postfix`
`micronaut.openapi.schema`

New configuration properties to replace deprecated:

`micronaut.openapi.schema.decorator.prefix`
`micronaut.openapi.schema.decorator.postfix`
`micronaut.openapi.schema.mapping`

### DIFF
--- a/openapi/openapi-custom-schema-for-class.properties
+++ b/openapi/openapi-custom-schema-for-class.properties
@@ -1,5 +1,5 @@
-micronaut.openapi.schema.io.micronaut.openapi.ObjectId=java.lang.String
-micronaut.openapi.schema.io.micronaut.openapi.JAXBElement=io.micronaut.openapi.MyJaxbElement
-micronaut.openapi.schema.io.micronaut.openapi.JAXBElement<test.XmlElement2>=io.micronaut.openapi.MyJaxbElement2
-micronaut.openapi.schema.io.micronaut.openapi.JAXBElement<test.XmlElement3>=io.micronaut.openapi.MyJaxbElement3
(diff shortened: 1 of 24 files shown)


sdelamo (commented):
This is a breaking change, is not it? Users upgrading with configuration keys will have to update their config.

altro3:
@sdelamo no. Now schemas with same name renamed with index, and after this improvement, default behaviour will be the same. But I add new mode `error`, if you want to fix schema duplication manually

altro3:
Ah... you talking about other options... So the main problem with this property: `micronaut.openapi.schema`. Need to change it, but how we can do it?

sdelamo:
> Ah... you talking about other options... So the main problem with this property: `micronaut.openapi.schema`. Need to change it, but how we can do it?

We will have to add additional configuration options without creating a breaking change.

altro3:
Ok, I'll fix it tomorrow

altro3:
@sdelamo fixed

PR #1773, opened 2024-09-17
fix: docs typo in endpoints servers definition

andrew-rosca:


graemerocher:
Thanks for the contribution!

PR #1777, opened 2024-09-20
Minor code improvements. Code format

altro3:


sdelamo (approved):
Not sure what prompted this refactor to avoid using Optional and the stream api.

altro3:
There is no point in creating any extra objects (Optional) just to do an if check.

PR #1780, opened 2024-09-20
Autogenerate Info block for openAPI spec

altro3:


sdelamo (approved):
Why do we need to add a default info block? Is not optional?

altro3:
Info block is required by OpenAPI specification. 

{877BC255-66CD-4BCA-B321-1D20D368E243}

altro3:
title and version are required in Info block

{E1418DDC-A0E9-405B-91AD-805A3A1EB338}

PR #1788, opened 2024-09-25
Add support generating Deprecated annotations and docs

altro3:
Removed support swagger1 annotations for java generator

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
@@ -44,6 +44,7 @@ import org.openapitools.codegen.CodegenProperty;
 import org.openapitools.codegen.CodegenResponse;
 import org.openapitools.codegen.IJsonSchemaValidationProperties;
 import org.openapitools.codegen.SupportingFile;
+import org.openapitools.codegen.VendorExtension;
 import org.openapitools.codegen.config.GlobalSettings;
 import org.openapitools.codegen.languages.AbstractJavaCodegen;
 import org.openapitools.codegen.languages.features.BeanValidationFeatures;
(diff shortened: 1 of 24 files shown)


openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
```diff
@@ -37,7 +37,7 @@ protected String generateFiles(MicronautCodeGenerator<?> codegen, String configP
         } catch (IOException e) {
             fail("Unable to create temporary directory for output");
         }
-        output.deleteOnExit();
+//        output.deleteOnExit();
```
sdelamo:
```suggestion
        output.deleteOnExit();
```

sdelamo (changes requested):
removing swagger 1 annotations is a breaking change. 

Why are we mixing this removal with a PR to add Deprecated annotations?

openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
altro3: fixed

altro3:
> removing swagger 1 annotations is a breaking change.
> 
> Why are we mixing this removal with a PR to add Deprecated annotations?
You are wrong, this change is not breaking. No one could generate code with swagger1 annotations before, simply because no one had the ability to change this setting (it is not passed in the plugin for gradle and maven).

In fact, this is just garbage that was left after moving the generator to this repository and this code was only in templates. By the way, in the generator for Kotlin, which I already wrote myself, there was no support for the first swagger.

I just removed the extra garbage, and this will not affect the previously generated code in any way.

graemerocher (approved):

graemerocher:
please resolve the conflicts

altro3:
@graemerocher done

graemerocher (approved):

PR #1789, opened 2024-09-26
Add support validation annotations extensions.

altro3:
Add processing custom messages to jakarta validation annotations with extensions:
`x-pattern-message`
`x-size-message`
`x-not-null-message`
`x-min-message`
`x-max-message`

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/Utils.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/Utils.java
@@ -156,20 +156,34 @@ public final class Utils {
             return result.toString();
         }
 
+        var patternMsg = (String) prop.vendorExtensions.get("x-pattern-message");
+        var sizeMsg = (String) prop.vendorExtensions.get("x-size-message");
+        var notNullMsg = (String) prop.vendorExtensions.get("x-not-null-message");
+        var minMsg = (String) prop.vendorExtensions.get("x-min-message");
+        var maxMsg = (String) prop.vendorExtensions.get("x-max-message");
+
         if (StringUtils.isNotEmpty(prop.pattern) && !prop.isDate && !prop.isDateTime) {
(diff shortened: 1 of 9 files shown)


sdelamo (commented):
Are are these  extensions documented anywhere` x-pattern-message`, `x-size-message`, `x-not-null-message`, `x-min-message`, `x-max-message` ? Is this something we are inventing for Micronaut OpenAPI or is it coming from a specification?

sdelamo:
can you merge 6.13.x ?

altro3:
This satisfies the specification because these are extensions.

The openapiGenerator core does not support these extensions, but for example x-pattern-message is supported by the spring generator: https://openapi-generator.tech/docs/generators/spring/

{E722B03A-BEF8-4070-8B4B-684781984953}

We are not obliged to support certain extensions, but we also have the right to create new extensions.

I added the rest of the extensions in connection with this PR: https://github.com/OpenAPITools/openapi-generator/pull/19630

In general, all these extensions are needed exclusively for more correct code generation. Support for new extensions does not affect the work of current users in any way, it simply adds the ability for them to describe the openapi specification so that the generated code is more accurate

sdelamo (commented):
Are these extensions defined or explained in our documentation? If not can you document, what we support?

altro3:
Rightly. I'll add documentation tomorrow

altro3:
Unfortunately, I don't have time to write full documentation yet. You see, at the moment we don't have any documentation for the generator at ALL. Therefore, we need to write it not only about these extensions, but about the entire generator as a whole.

Therefore, I suggest not to bother with describing the generator features for now. I mean, specifically in these PRs, but in the future it will be time to write documentation for the generator. I was going to do this, but I haven't found the time yet

graemerocher:
Please resolve the conflicts

graemerocher (approved):

PR #1799, opened 2024-09-30
Correct generating no vars classes

altro3:


altro3:
Don't know how to create schema class without properites... But it possibly with kotlin noargs plugin. See details here: https://github.com/OpenAPITools/openapi-generator/issues/14710

sdelamo (dismissed):
Can you add a test for this?

altro3:
The problem is that I can't figure out in which case a class without attributes is generated. But it is definitely possible, since there is an issue. I checked it locally by forcing noArgs = true into the calculations

altro3:
@sdelamo Tests added

graemerocher (approved):

graemerocher:
Please resolve conflicts

altro3:
@graemerocher done

PR #1857, opened 2024-11-08
Update views versions

altro3:
OpenAPI Explorer 2.2.732
Rapidoc 9.3.8
ReDoc 2.2.0
Swagger UI 5.18.2


sdelamo (commented):
@altro3 can you please add links to where these files are being copied from? Also which licenses these files have. 

It is hard to review a PR with minimized javascript files.

altro3:
OpenApiExplorer: https://unpkg.com/openapi-explorer@2.2.730/dist/browser/openapi-explorer.min.js
Sources repo: https://github.com/Authress-Engineering/openapi-explorer

Rapidoc: https://unpkg.com/rapidoc@9.3.6/dist/rapidoc-min.js
Sources repo: https://github.com/rapi-doc/RapiDoc

Redoc: https://unpkg.com/redoc@2.2.0/bundles/redoc.standalone.js
Sources repo: https://github.com/Redocly/ReDoc/

Swagger UI: https://unpkg.com/swagger-ui@5.18.2/dist/swagger-ui.js
Sources repo: https://github.com/swagger-api/swagger-ui

Where I need to add licenses?

And why we need add licenses only for these libs? Then we should add licenses for each of the libraries we use. But I don't think it makes sense, because micronaut is not a commercial product.

sdelamo:
> OpenApiExplorer: https://unpkg.com/openapi-explorer@2.2.730/dist/browser/openapi-explorer.min.js Sources repo: https://github.com/Authress-Engineering/openapi-explorer
> 
> Rapidoc: https://unpkg.com/rapidoc@9.3.6/dist/rapidoc-min.js Sources repo: https://github.com/rapi-doc/RapiDoc
> 
> Redoc: https://unpkg.com/redoc@2.2.0/bundles/redoc.standalone.js Sources repo: https://github.com/Redocly/ReDoc/
> 
> Swagger UI: https://unpkg.com/swagger-ui@5.18.2/dist/swagger-ui.js Sources repo: https://github.com/swagger-api/swagger-ui
> 
> Where I need to add licenses?
> 
> And why we need add licenses only for these libs? Then we should add licenses for each of the libraries we use. But I don't think it makes sense, because micronaut is not a commercial product.

Can you create README.md files in the foldes where these files are. In the contents of the readme.md file point to the original source. Also, mention the license of the libraries in those README.md files. To make sure we are allowed to copy those files into a different open source library and redistributed it.

altro3:
Do we need these files in final jar file? Or, maybe I can add it to documentation?

altro3:
And still, I don't get it. Why should we add links to licenses of only some open source libraries? Why haven't we added licenses for jackson , freemarker , adoc , etc. What's the difference between javascript scripts and java ?

sdelamo:
> And still, I don't get it. Why should we add links to licenses of only some open source libraries? Why haven't we added licenses for jackson , freemarker , adoc , etc. What's the difference between javascript scripts and java ?

[I have added  README.md files with what I meant.](https://github.com/micronaut-projects/micronaut-openapi/pull/1857/commits/150de6aa4f2deb83f069048bc95295fd7ccf65eb). We need to document where these files come from to ease the update by anyone and we need to document what is the license in the original repository. I want to be sure that we can legally copy them and bundle them in `micronaut-openapi`. I have also written a bash script that downloads the latest versions.[ I have run the bash script and updated the JS files](https://github.com/micronaut-projects/micronaut-openapi/pull/1857/commits/bf452d01be4377c7bf6b9712d5a40affcaa471a2). 

I would prefer if we did not include minimized js files in our jar. We could just use an absolute link to unpkg.com. That way, it will be less work to maintain and less risk of shipping a security vulnerability in a minimized js file.

sdelamo (approved):

altro3:
After your changes, will the scripts be included in the final jar file or not?

I will explain why it was done this way: so that when using our library, developers would get a full-fledged solution without external resources. I remember that a couple of years ago there was a request for this and I released such a solution. It would be good if after your changes with downloading scripts, the final archive did not change

sdelamo:
> After your changes, will the scripts be included in the final jar file or not?

Yes. 

> I will explain why it was done this way: so that when using our library, developers would get a full-fledged solution without external resources. I remember that a couple of years ago there was a request for this and I released such a solution. 

Yes, I understand the advantage. However, I think the disadvantages are not worth for us. We are now shipping minimised javascript into one of our artifacts. These JS files are hard/impossible to review and may contain vulnerabilities .  

> It would be good if after your changes with downloading scripts, the final archive did not change. 

Sorry, I don't understand what do you mean. I downloaded the js files with the bash script. Do you think it is is incorrect?

altro3:
I haven't tested this solution yet, the main thing is that if I run the build locally on Windows, these files are also added to the archive. I often publish to a local repository to test my fixes.

altro3:
Ok, I understand your solution. You suggest running a script to download scripts, rather than running around different sites and copying these crypts. I don't argue that it's more convenient.

Then the question is: how does this solution differ from manual work? Yes, it's better, but it will still lead to the appearance of PR with modified minified javascript files.

altro3:
@sdelamo Ok, I checked it with windows again. So it works fine with gitbash or WSL on Windows. I agree that I like this method of updating scripts better.

sdelamo:
> Then the question is: how does this solution differ from manual work? Yes, it's better, but it will still lead to the appearance of PR with modified minified javascript files.


someone reviewing the PR could run the scripts locally and at least try the files are the same.

altro3:
Yes, I already understood. I agree that it is more convenient with a script.

PR #1983, opened 2025-02-12
Added new option `jsonIncludeAlwaysForRequiredFields` to the generator to add annotation `@JsonInclude(ALWAYS)` for required fields

### LINKED ISSUE #1979: Generated code does not observe required properties
### Expected Behavior

When an openApi spec contains a schema with a required property, I expect that a response always contains this property, even if it is an empty String. This equates to generating `@JsonInclude(value = JsonInclude.Include.ALWAYS)` on the respective model java class.

### Actual Behaviour

The generated model class does not have the annotation `@JsonInclude(value = JsonInclude.Include.ALWAYS)` on the member and hence the property is omitted in the response body.

I saw some code in the [pojo.mustache](https://github.com/micronaut-projects/micronaut-openapi/blob/24d631cedfa62af517647ed8c9e083f1c67ec0b0/openapi-generator/src/main/resources/templates/java-micronaut/common/model/pojo.mustache#L4) that seemed to observe this for maps, but not for other properties (not sure, don't know mustache).

### Steps To Reproduce

See the attached project and run the TextsTests (1 will fail). The issue with the timestamp in that test is something else, you can ignore it.

[micronaut-openapi.zip](https://github.com/user-attachments/files/18750708/micronaut-openapi.zip)

### Environment Information

- MacOS 15.3
- JDK 21
- Micronaut Openapi 4.4.4

### Example Application

https://github.com/user-attachments/files/18750708/micronaut-openapi.zip

### Version

4.7.4

altro3:
Fixed #1979

### DIFF
--- a/openapi-generator/src/main/resources/templates/java-micronaut/common/model/jackson_annotations.mustache
+++ b/openapi-generator/src/main/resources/templates/java-micronaut/common/model/jackson_annotations.mustache
@@ -9,14 +9,21 @@
         {{#items.isNullable}}
     @JsonInclude(content = JsonInclude.Include.ALWAYS{{^required}}, value = JsonInclude.Include.USE_DEFAULTS{{/required}})
         {{/items.isNullable}}
-        {{^items.isNullable}}{{^required}}
+        {{^items.isNullable}}
+            {{^required}}
     @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
-        {{/required}}{{/items.isNullable}}
+            {{/required}}
+        {{/items.isNullable}}
     {{/isMap}}
     {{^isMap}}
         {{#required}}
             {{#isReadOnly}}
     @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
+            {{/isReadOnly}}
+            {{^isReadOnly}}
+                {{#jsonIncludeAlwaysForRequiredFields}}
+    @JsonInclude(JsonInclude.Include.ALWAYS)
+                {{/jsonIncludeAlwaysForRequiredFields}}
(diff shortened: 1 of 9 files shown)


openapi-generator/src/main/resources/templates/java-micronaut/common/model/jackson_annotations.mustache
```diff
@@ -9,15 +9,20 @@
...
-        {{/required}}{{/items.isNullable}}
+            {{/required}}
+        {{/items.isNullable}}
     {{/isMap}}
     {{^isMap}}
         {{#required}}
             {{#isReadOnly}}
     @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
             {{/isReadOnly}}
+            {{^isReadOnly}}
```
jakobmagiera, line 23: Can you help me understand why this only applies to non-readonly props?

altro3: Realy don't know, it is old code template. If you think this is wrong - I can try to change it

altro3:
@jakobmagiera so, I think it's better to add this annotation optionally. I've added new option `jsonIncludeAlwaysForRequiredFields`. Set it `true` and annotation `@JsonInclude(ALWAYS)` will be added

altro3:
@graemerocher Also this, pls

sdelamo (approved):

sdelamo:
@altro3 can you resolve the conflicts?

jakobmagiera (commented):
Thanks - tested successfully with the new release! :)

PR #1999, opened 2025-02-22
Fixed OOM and Stream closed for openapi-generator

### LINKED ISSUE #1997: Class generation fails when there are multiple OpenAPI files in the project
### Expected Behavior

As a user, I want to generate classes for multiple OpenAPI specification files

### Actual Behaviour

It often happens that the generation process ends with an error: Stream closed. The problem occurs randomly. Out of 10 attempts, 8 can fail. Sometimes the process completes successfully, and sometimes an exception is thrown. 

I believe that the frequency of the error depends on the size of the files – in my sample application, it occurs less frequently than with 'serious' specifications

In the case of large files, there is also an OutOfMemory problem. `java.lang.OutOfMemoryError: Metaspace`

### Steps To Reproduce

1. Create a simple micronaut application
2. Add multiple OpenApi spec files.
3. Configure the plugin
```
    openapi {
        server("serverOne", file("src/main/resources/openapi/my-api-first.yml")) {
            apiPackageName = "com.example.first.api"
            modelPackageName = "com.example.first.model"
            controllerPackage = "com.example.first.controller"
            useReactive = false
            useOptional = true
            generateHttpResponseWhereRequired = true
        }
        server("serverTwo", file("src/main/resources/openapi/my-api-second.yml")) {
            apiPackageName = "com.example.second.api"
            modelPackageName = "com.example.second.model"
            controllerPackage = "com.example.second.controller"
            useReactive = false
            useOptional = true
            generateHttpResponseWhereRequired = true
        }
        server("serverThree", file("src/main/resources/openapi/my-api-third.yml")) {
            apiPackageName = "com.example.third.api"
            modelPackageName = "com.example.third.model"
            controllerPackage = "com.example.third.controller"
            useReactive = false
            useOptional = true
            generateHttpResponseWhereRequired = true
        }
    }
```
4. Run `./gradlew clean compileJava`

[error-stream-closed.txt](https://github.com/user-attachments/files/18884444/error-stream-closed.txt)

### Environment Information

- OpenSuse Tumbleweed
- JDK23
- Micronaut Plugin 4.4.5 / OpenApi plugin: 6.13.1 (but update to 6.13.2 didnt help)

### Example Application

https://github.com/matadini/openapi-micronaut-multiple-spec-files-bug

### Version

4.7.6

altro3:
Disable generating openapi-generator metadata.

Fixed #1997
Fixed #https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/1070

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/MicronautCodeGeneratorEntryPoint.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/MicronautCodeGeneratorEntryPoint.java
@@ -109,6 +109,7 @@ public final class MicronautCodeGeneratorEntryPoint {
 
         // Generate
         var generator = new DefaultGenerator();
+        generator.setGenerateMetadata(false);
         for (OutputKind outputKind : OutputKind.values()) {
             generator.setGeneratorPropertyDefault(outputKind.generatorProperty, "false");
         }


altro3:
@sdelamo @graemerocher Can't write test for it, because this problem happens whe we run many gradle tasks for generating client classes in parallel mode with gradle plugin. Tested it locally.

The problem did not always occur, apparently depending on the number of parallel threads. The main problem was in the generation of metadata in the generator itself, since the same file was trying to be generated in parallel processes.

I also disabled unnecessary `logback.xml` and `application.yml` for client generation

melix (commented):
Thanks @altro3 . I'm not quite confident about these changes. It's unclear to me why they fix the stream closed issue. Also removing the supporting files seems like a breaking change (actually you also add more to the server codegen). This is probably due to my lack of knowledge of how this works, but I would appreciate a deeper explanation of the changes in that PR before approving. Thank you!

altro3:
@melix That's the main fix:

{462DF243-CA56-484D-ACAF-576249E4BE95}

The problem inside openApi generator library, if you create metadata file (we don't need it). By default this setting is true and OOM and `Stream closed` exception was with creating this file. I just disable generating of this file, because we don't need it.

About `application.yml` and `logback.xml` - ok, if you think this is breaking chnage - I reverted these changes

altro3:
@melix ping

melix:
> The problem inside openApi generator library, if you create metadata file (we don't need it). By default this setting is true and OOM and Stream closed exception was with creating this file.

So you are saying there's a bug upstream, in which case this PR would be a workaround, not really a fix? What is this metadata about and who is "we"? I'm kind of hesitating on approving the changes since for example [this change](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1052/files#diff-0d8feb3d2e60a4085c22391e83952def203f6d69a5ae209e04b1886d52a971f9R126) was a breaking change which led to builds failing after upgrading to a minor release (not blaming you, I should have caught it).

altro3:
Yes, this is a bug in openapi generator. But this is not a real bug. This problem occurs in the script of the plugin from micronaut, but it does not occur in the plugin from openapi generator . So this is not even a bug, but rather the specifics of the use of this library by the micronaut team. They don't have this bug because they use their own gradle plugin. And the problem arises precisely because of the specifics of parallelizing the creation of API and models. I don't see any reason to doubt the correctness of this solution, because metainformation is a file (without extension) containing information about which version of openapi generator was used. So fears that something will break are in vain

{C4292B85-08FE-458E-BEB4-FFD5F771E439}

PS Ok, not "we", just a `micronaut team`

melix:
Ok thanks. It's strange that there's a parallelism bug, this would sound like shared state. In this case we might want to isolate generation, or maybe use a lock if it's not possible.

melix (approved):

PR #2007, opened 2025-02-26
Gradle 8.13

altro3:


graemerocher:
I believe there is a change / regression in Gradle 8.13 that requires an update to our internal Gradle plugins

altro3:
As I understand, it's already fixed and works correctly by this:

https://github.com/micronaut-projects/micronaut-build/pull/754
https://github.com/micronaut-projects/micronaut-openapi/pull/2006

graemerocher (approved):

PR #2069, opened 2025-04-02
Added support multiple documents for Scalar

altro3:
Scalar 1.28.11
Swagger UI 5.20.3


graemerocher:
is there a way we can pull the scalar JS files from CDN? Having huge JS files into this repo makes it harder to review changes

altro3:
@graemerocher That's what we do. We just run the bash script and it downloads all the js files from the CDN - I don't download anything manually.

Look at this file: https://github.com/micronaut-projects/micronaut-openapi/blob/6.15.x/download-js-files.sh

I run it to update the UI scripts

graemerocher:
ok but why isn't that file part of the build and why are we checking into source control these downloaded files?

altro3:
I think it's historically been like that. Scripts used to be loaded from external sources, and a couple of years ago users really asked to localize scripts to avoid external requests - I did it, but the build process didn't change. If you know how to improve this process and not store files in git - that's great, but I don't know how to do it

graemerocher:
@melix do you know if we can do this in the Gradle build? Makes me uncomfortable from a security perspective maintaining this javascript code in this repository.

melix:
That should definitely be done as part of the build, there's no reason not to do it.

altro3:
In fact, I don't quite understand how you propose to integrate this into the build process. Currently, the files are stored in Git, but this eliminates unexpected problems. That is, if you start loading these files with each build, then you will either have to strictly tie to the version of each of the scripts, or the latest version of the script will always be loaded, which can sometimes lead to problems. Unlikely, but it can.

What I mean is that the current solution, although not the most beautiful, there are definitely no problems with it because the script versions are updated manually, not automatically.

graemerocher:
they should be versioned like any other artefact. In fact if we can get vulnerability checkers to trigger for them even better.

@melix when you get a minute can you look into this?

melix:
Yes, these dependencies should be fetched at build time like any other dependency. The difference is that we're in JS world, so a fixed version is not... fixed in time!

melix:
So I think we can merge this PR independently of the rework of how these libraries are fetched from CDN. I will work on a separate PR clean that up.

altro3:
agree

graemerocher (approved):

PR #2073, opened 2025-04-04
Use a Gradle task to download external JS dependencies

melix:
@graemerocher this is the follow up to #2069


gradle/libs.versions.toml
```diff
@@ -44,6 +44,14 @@ micronaut-spring = "5.10.0"
 micronaut-docs = "2.0.0"
 sonatype-scan = "3.0.0"
 
+# javascript libraries
+js-swagger-ui="5.20.5"
+js-openapi-explorer="2.2.733"
+js-redoc="2.4.0"
+js-rapidoc="9.3.8"
+js-rapipdf="2.2.1"
+js-scalar="1.28.11"
```
graemerocher, line 53: is there a way to get dependency scanners to pick this up?

graemerocher (approved):

gradle/libs.versions.toml
melix: I think one option would be to create take nodejs lock files or somethinig. Not very familiar with this.

altro3 (approved):

altro3:
@melix Forgot to ask you: why you change CDN to NPM ? NPM not so fast, as CDN, but CDN support versioning too. Like this: https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.28.12/dist/browser/standalone.js

altro3:
@melix sorry, I'm blind. You didn't change it

PR #2077, opened 2025-04-08
Improvements for swagger 2.2.32

altro3:


openapi-common/src/main/java/io/micronaut/openapi/SimpleSchema.java
```diff
@@ -1,61 +0,0 @@
...
-import java.util.Objects;
-
-import io.swagger.v3.oas.models.media.Schema;
-
-/**
- * Copy of MapSchema but without type 'object'. Need this class to correct deserializing schema without type.
- *
- * @since 4.8.7
- */
-public class SimpleSchema extends Schema<Object> {
```
sdelamo, line 27: deleting a public class is a breaking change.

openapi-common/src/main/java/io/micronaut/openapi/swagger/core/util/PrimitiveType.java
```diff
@@ -505,6 +622,12 @@ public String getCommonName() {
 
     public abstract Schema<?> createProperty();
 
+    public abstract Schema<?> createProperty31();
```
sdelamo, line 625: why is this method named `createProperty31`? I assume this is for OpenAPI 3.1? I don't think we should support multiple versions. As far as I see there is not an OpenAPI 3.2 finalized specification. I don't think we should move to OpenAPI 3.2. until that it is ready. https://github.com/OAI/OpenAPI-Specification/tags

sdelamo (changes requested):

openapi-common/src/main/java/io/micronaut/openapi/SimpleSchema.java
altro3: Hm, This class I created, because early swagger haven't it, but now they added `AbribtratrySchema` which the same with my old `SimpleSchema` - object to serialization eqauls with the `Schema`, but, without `type` field. What do you suggest? We don't need this class anymore

openapi-common/src/main/java/io/micronaut/openapi/swagger/core/util/PrimitiveType.java
altro3: This class is copy of the same class from `swagger-core` with some my improvements. This method from swagger `2.2.30`.

And we already support OpenAPI 3.1

sdelamo: I know we support Open API 3.1. What is this PR solving? There are not tests in this PR.

openapi-common/src/main/java/io/micronaut/openapi/SimpleSchema.java
sdelamo: annotate it with `@Deprecated(forRemoval=true` and say in the javadoc is no longer used.

openapi-common/src/main/java/io/micronaut/openapi/swagger/core/util/PrimitiveType.java
altro3: That PR not solve anything, just more correct integration with swagger 2.2.30

openapi-common/src/main/java/io/micronaut/openapi/SimpleSchema.java
altro3: done

altro3:
Now this PR haven't any breaking changes.

Why You create branch 6.16.0? My changes not contains any major changes, or major librariies versions. Are you suggesting that users wait for half a year until Microsoft 4.9.0 is released and not use my fixes? Because of this system, people wait for fixes for several months, although there is no point in doing so at all

openapi-common/src/main/java/io/micronaut/openapi/SimpleSchema.java
graemerocher: most of this logic was not on the public API since the openapi module is an annotation processor so not sure if this was necessary

test-suite-java-spring/src/main/java/io/micronaut/openapi/spring/api/MyController.java
```diff
@@ -0,0 +1,83 @@
...
+import static io.swagger.v3.oas.annotations.enums.ParameterIn.QUERY;
+
+// end::imports[]
+// tag::clazz[]
+@RestController
+@RequestMapping("/my-api")
+public class MyController {
+
+    @PostMapping(value = "/testMultipart/{pathVar}", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
+    public void testMultipart(@RequestBean WrappedRq rq) {
```
graemerocher: not sure that forcing spring users to use a Micronaut specific feature makes sense? Why is this being recommended in the docs?

openapi-common/src/main/java/io/micronaut/openapi/swagger/core/util/ModelDeserializer.java
```diff
@@ -56,7 +55,17 @@
  */
 public class ModelDeserializer extends JsonDeserializer<Schema> {
 
-    protected boolean openapi31;
+    static Boolean useArbitrarySchema = false;
+
+    static {
+        if (System.getenv(Schema.USE_ARBITRARY_SCHEMA_PROPERTY) != null) {
```
graemerocher, line 61: why are we depending on environment variables here? Annotation processors should not use env variables this way.

graemerocher (changes requested):

openapi-common/src/main/java/io/micronaut/openapi/swagger/core/util/ModelDeserializer.java
altro3: That's not our env vars. This code from `swagger-core`

test-suite-java-spring/src/main/java/io/micronaut/openapi/spring/api/MyController.java
altro3: This is necessary so that micronaut openapi understands that there is a wrapper class parameter in front of it, and inside there can be both parts of a multipart request, and path variables, and variables from the request.

Spring does not have a special annotation to parse everything correctly - it does this without any additional annotations, unlike micronaut. But openapi cannot understand what kind of parameter is in front of it, so it needs to be marked somehow for correct processing. In order not to create an additional annotation, I suggest using the existing RequestBean annotation.

If you look at springdoc - they encountered the same problem and were forced to add their own annotation to mark such parameters - `@ParameterObject`

graemerocher (approved):

altro3:
@sdelamo please, aproove this PR

sdelamo (approved):

PR #2122, opened 2025-05-12
Add ability to interpret primary constructor arguments without annotations as not required properties

### LINKED ISSUE #2116: Property should be marked as Nullable by default
### Expected Behavior

In Java, if not marked otherwise, a property is nullable by default.
When my property in my class has no annotation, I expect it to be marked as nullabled in the openapi specification file generated.

### Actual Behaviour

When my property in my class has no annotation, it is marked as not null.

### Steps To Reproduce

- Create a Java Project, with a class that has nullable property (but not explicitly marked)
- Generate the openapi file
- Property is marked required

### Environment Information

JDK 21

### Example Application

https://github.com/kestra-io/kestra

### Version

6.15

altro3:
By config property `micronaut.openapi.constructor-arguments-as-required`

Fixed #2116


altro3:
@graemerocher @sdelamo I don't know why they initially did the opposite, that arguments in the primary constructor are mandatory. In general, this is not a correct interpretation, because the presence of an argument in the constructor does not make it mandatory, unless an explicit `@NonNull` annotation is specified.

But to make sure no one breaks anything, I added the ability to softly switch the interpretation of arguments without annotations for those who use this feature.

anna-geller:
@altro3 big thanks for the PR, this will also solve the issue on our side at Kestra.

Hi @graemerocher @sdelamo, could you review the PR sometime soon-ish or ask some other maintainers? We're also blocked due to this issue. We're working on an SDK and due to this bug, our models are not correctly generated. We'd kindly appreciate it if you could get that reviewed and merged. Thanks so much in advance!

altro3:
@graemerocher ping

graemerocher (approved):

PR #2131, opened 2025-05-16
Fix `resultSchemaName` not checking all existing class entries

### LINKED ISSUE #2130: SchemaDefinitionUtils: `schemaNameToClassNameMap` doesn't take into account `shemaNameSuffixCounterMap` resulting in duplicated schemas being generated
### Expected Behavior

When generating an openapi spec with two classes of the same name but in different packages; e.g. `one.Request` and `two.Request`, I would expect each `Request` type to be contained within a single generated schema; e.g. `#/components/schemas/Request` and `#/components/schemas/Request_1` following the current naming conventions in this code.

### Actual Behaviour

The first encountered `Request` class will have a single schema generated for it, but the second encountered `Request` class will have a schema generated for each time it is referenced.  For example if both `one.Request` and `two.Request` are referenced 10 times each and `one.Request` was encountered first, then there will be one schema generated for the `one.Request` named `#/components/schemas/Request` and 10 schemas generated for `two.Request` named `#/components/schemas/Request_1`, `#/components/schemas/Request_2`, ..., `#/components/schemas/Request_10`

This is due to how the `schemaNameToClassNameMap` doesn't take into account entries in the `shemaNameSuffixCounterMap` (side-note: this property is incorrectly named in the code, and so I've incorrectly named it here as well.  The 'c' is missing from schema).

When the first `Request` (from `one.Request`) is encountered an entry of `Request: one.Request` will be added to `schemaNameToClassNameMap`. The next time `one.Request` is encountered, the `schemaNameToClassNameMap` map will return the package `one.Request`, which matches the requested pacakge and therefore the schema-name `Request` will be used.

However, when the first request for `two.Request` is encountered, the entry in `schemaNameToClassNameMap` returns a different package which does not match this `Request` package (`one.Request` != `two.Request`). This results in an entry being made in the `shemaNameSuffixCounterMap` with the value of `Request: 1`, which causes this schema to be renamed to `Request_1` and then stored in the `schemaNameToClassNameMap` as `Request_1: two.Request`. When the next `two.Request` comes in, only the `Request` value is checked in `schemaNameToClassNameMap`, resulting in a package mismatch again. This will now update `shemaNameSuffixCounterMap` to the value `Request: 2` and a new entry is added to `schemaNameToClassNameMap` as `Request_2: two.Request`.  This repeats for each reference to `two.Request` as the code doesn't know it should have also checked `schemaNameToClassNameMap` for the entry `Request_1`, which would match the `two.Request` package.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

6.15.0

colesnodgrass:
- fix #2130 
- fix typo - `shemaNameSuffixCounterMap` -> `schemaNameSuffixCounterMap`
- extract `resultSchemaName` to new `computeResultSchemaName` function
    - if this is the first time this schema has been encountered (e.g. `foo.Data`)
        - add an entry to `schemaNameToClassNameMap` (e.g. `Data : foo.Data`)
    - if this is schema already exists _and_ the packages match
        - use existing `resultSchemaName` (e.g. `Data`)
     - if this schema already exists _and_ the packages do not match
         - determine how many additional schemas may exists (the value stored in `schemaNameSuffixCounterMap`)
             - if no additional entries, create an entry in `schemaNameSuffixCounterMap` (e.g. `Data : 1`) and `schemaNameToClassNameMap` (e.g. `Data_1 : foo.Data`).
         - check each of these values for a package match
             - if found, return the existing `resultSchemaName` (e.g. `Data_1`)
         - if no match is found, increment the entry in the `schemaNameSuffixCounterMap` by 1 (e.g. `Data : 2`), add an entry to `schemaNameToClassNameMap` with this new value (e.g. `Data_2 : foo.Data`)

Fixed #2130

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
@@ -258,7 +258,7 @@ public final class SchemaDefinitionUtils {
     /**
      * Stores class name counters for schema suffix, when found classes with same name in different packages.
      */
-    private static Map<String, Integer> shemaNameSuffixCounterMap = new HashMap<>();
+    private static Map<String, Integer> schemaNameSuffixCounterMap = new HashMap<>();
     /**
      * {@link PropertyNamingStrategy} instances cache.
      */
@@ -273,7 +273,7 @@ public final class SchemaDefinitionUtils {
     public static void clean() {
         inProgressSchemas = new ArrayList<>(10);
         schemaNameToClassNameMap = new HashMap<>();
-        shemaNameSuffixCounterMap = new HashMap<>();
+        schemaNameSuffixCounterMap = new HashMap<>();
         propertyNamingStrategyInstances = new HashMap<>();
     }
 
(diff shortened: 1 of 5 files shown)


altro3:
@colesnodgrass Thanks for PR! Could you add a test?

colesnodgrass:
@altro3 Sure thing.  I would have included one in this PR but the `SchemaDefinitionUtils.java` file doesn't have any existing tests.  Do you want me to add tests for this file or is there a different place where this test should be created?

altro3:
Look to existing tests, it doesn't matter how you will do it. You can create different file with your tests, but these tests should not be for SchemaDefinitionUtils, they should reproduce your classes structure and the problem what you fixed

colesnodgrass:
@altro3 test added

altro3 (approved):

altro3:
I merged it, but more correct test shoul be in module `openapi` and  looks like this (to the future):
```groovy
    void "test Schema names for classes with same name in different packages"() {

        buildBeanDefinition('test.MyBean', '''
package test;

import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;
import io.micronaut.http.annotation.Produces;
import io.micronaut.openapi.model.one.Response;

@Controller("/model")
class ModelController {

    @Get("/model-one")
    @Produces("application/json")
    public Response getResponseOne() {
        return new Response("one");
    }

    @Get("/model-one-b")
    @Produces("application/json")
    public Response getResponseOneB() {
        return new Response("one-b");
    }

    @Get("/model-two")
    @Produces("application/json")
    public io.micronaut.openapi.model.two.Response getResponseTwo() {
        return new io.micronaut.openapi.model.two.Response("two");
    }

    @Get("/model-two-b")
    @Produces("application/json")
    public io.micronaut.openapi.model.two.Response getResponseTwoB() {
        return new io.micronaut.openapi.model.two.Response("two-b");
    }

    @Get("/model-two-c")
    @Produces("application/json")
    public io.micronaut.openapi.model.two.Response getResponseTwoC() {
        return new io.micronaut.openapi.model.two.Response("two-c");
    }
}

@jakarta.inject.Singleton
class MyBean {}
''')
        when:
        var openApi = Utils.testReference

        then:
        openApi

        openApi.paths."/model/model-one".get.responses."200".content."application/json".schema.$ref == "#/components/schemas/Response"
        openApi.paths."/model/model-one-b".get.responses."200".content."application/json".schema.$ref == "#/components/schemas/Response"
        openApi.paths."/model/model-two".get.responses."200".content."application/json".schema.$ref == "#/components/schemas/Response_1"
        openApi.paths."/model/model-two-b".get.responses."200".content."application/json".schema.$ref == "#/components/schemas/Response_1"
        openApi.paths."/model/model-two-c".get.responses."200".content."application/json".schema.$ref == "#/components/schemas/Response_1"

        openApi.components.schemas.size() == 2
        openApi.components.schemas.containsKey("Response")
        openApi.components.schemas.containsKey("Response_1")
    }
```

PR #2132, opened 2025-05-17
Improvements for authorization classes generation.

### LINKED ISSUE #2115: Gradle generation of two clients with enabled auth cause errors in runtime
### Expected Behavior

Both clients connect and work as described in the yml file.

### Actual Behaviour

I constantly get a 400 error. The error seems to be related to the fact that the request is handled by two filters. The first handles the request correctly, but the second tries to use a different login method (OAuth), which is not configured. 

The problem can be caused by a fact that a filed in AuthorizationBinder 
AUTHORIZATION_NAMES = "micronaut.security.AUTHORIZATION_NAMES";
is not prefix/sufix by invokerPackage
Ie

"micronaut.security.{{invokerPackage}}.AUTHORIZATION_NAMES";

And both filters can request the same attributes (is it a leak ? )

In addition, is there any way to quickly replace templates? I tried several dirs in resources but they are not working. 

### Steps To Reproduce

1. Take an example from the open api client generator guide. 
2. Add another API description. 
3. Add new client configuration to gradle 
3. Turn on auth parameter in both configurations
4 Set parameters apiPackageName, invokerPackageName, basePathSeparator, modelPackageName separately for each model. 
5. Try to build 

### Environment Information

-JDK 21
-MacOS

### Example Application

_No response_

### Version

4.6.1

altro3:
Fixed #2115


altro3:
So, what has changed:

1. Now, in order for you to have no problems with filters and request attributes, just set the `clientId` in the generator settings. After that, the filter will be applied ONLY for requests of this client. Just in case, the attribute also now includes the clientId . In addition, the `HttpBasicAuthConfig` and `ApiKeyAuthConfig` classes are now also tied to the clientId and look like this: `@EachProperty("security.<clientId>.api-key-auth")` and `@EachProperty("security.<clientId>.basic-auth")`

The final configuration will look something like this:
```groovy
    openapi {
        version = ver.openapi
        client("client1", file("swagger.yml")) {
            clientId = "client1"
            apiPackageName = "com.generated.client1.api"
            modelPackageName = "com.generated.client1.model"
            invokerPackageName = "com.generated.client1.auth"
            useAuth = true
        }
        client("client2", file("swagger2.yml")) {
            clientId = "client2"
            authorizationFilterPattern = "/test;/different/simple/path/**;/something/else"
            apiPackageName = "com.generated.client2.api"
            modelPackageName = "com.generated.client2.model"
            invokerPackageName = "com.generated.client2.auth"
            useAuth = true
        }
    }
```
2. The second method, more complicated: use a single configuration and a single filter for all clients:

To do this, generate the configuration and filter only in one of the clients, in all the others, disable the generation of Authorization classes. Next, set a single invockerPackage package in all clients.

This way, the configuration and filter will be in a single instance, and in the other clients, only the @Authorization annotation will be set above the methods in the client.

The final configuration will look something like this:
```groovy
    openapi {
        version = ver.openapi
        client("client", file("swagger.yml")) {
            apiPackageName = "com.generated.client1.api"
            modelPackageName = "com.generated.client1.model"
            invokerPackageName = "com.generated.auth"
            useAuth = true
        }
        client("client2", file("swagger2.yml")) {
            authorizationFilterPattern = "/test;/different/simple/path/**;/something/else"
            apiPackageName = "com.generated.client2.api"
            modelPackageName = "com.generated.client2.model"
            invokerPackageName = "com.generated.auth"
            useAuth = true
            additionalProperties = [
                    generateAuthClasses: false
            ]
        }
    }
```
As you see `invokerPackageName` the same for both clients

3. The third method: use a configuration for set of clients and a single filter for selected clients:

To do this, generate the configuration and filter only in one of the clients, in all the others, disable the generation of Authorization classes. Next, set a single invockerPackage package in all clients.

This way, the configuration and filter will be in a single instance, and in the other clients, only the @Authorization annotation will be set above the methods in the client.

The final configuration will look something like this:
```groovy
    openapi {
        version = ver.openapi
        client("client", file("swagger.yml")) {
            clientId = "client1"
            apiPackageName = "com.generated.client1.api"
            modelPackageName = "com.generated.client1.model"
            invokerPackageName = "com.generated.auth"
            useAuth = true
            additionalProperties = [
                    authFilterClientIds : ["client1", "client2"]
            ]
        }
        client("client2", file("swagger2.yml")) {
            clientId = "client2"
            authorizationFilterPattern = "/test;/different/simple/path/**;/something/else"
            apiPackageName = "com.generated.client2.api"
            modelPackageName = "com.generated.client2.model"
            invokerPackageName = "com.generated.auth"
            useAuth = true
            additionalProperties = [
                    generateAuthClasses: false
            ]
        }
    }
```

Also I've added some helpfull properties:

```properties
    useOauth = true // Generate AuthorizationFilter with support OAuth2.0 or not.
    useBasicAuth = true // Generate HttpBasicAuthConfig class or not
    useApiKeyAuth = true // Generate ApiKeyAuthConfigclass or not
    authFilter = true // Generate AuthorizationFilter or not
    generateAuthClasses = true // Generate authorization classes or not. If "false" annotaion will be added only in Client class, but you need to create annotation class yourself

    authConfigName = test // This value will be used in generated configuration path and `AUTHORIZATION_NAMES`  request attribute. This is useful, if you use this AuthFilter for several clients

    authFilterClientIds = client1;client2 // List of client IDs for which this filter will be applied. This is useful if you use multiple clients to different internal microservices that share common authentication, such as an API KEY.
    authFilterExcludedClientIds = "client3", "client4" // List client IDs for which this filter will NOT be applied
    authorizationFilterPatternStyle = regex // Pattern style for Authorization filter. Available options: ANT (default), REGEX
```

graemerocher (approved):

PR #2142, opened 2025-05-20
Fix body unwrapped parameter

### LINKED ISSUE #2136: Extending interface changes generated spec on @Body annotation with value member present
### Expected Behavior

Given this interface:
```java
@OpenAPIGroup(value = {"Child"})
@Controller
public interface Child extends Parent {
  @Post
  void doSomething(@Body("a") String a);
}

interface Parent {}
```

I would expect the generated schema of the request body in `Child.yml` to look like this (as it is working without the inheritance present):
```yaml
paths:
  /:
    post:
      operationId: doSomething
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                a:
                  type: string
```

### Actual Behaviour

However, it is generated utilizing a `oneOf` with two identical options:
```yaml
  /:
    post:
      operationId: doSomething
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
              - type: object
                properties:
                  a:
                    type: string
              - type: object
                properties:
                  a:
                    type: string
            example: null
```

### Steps To Reproduce

1. Create an interface
2. Add an operation which references a specific body field instead of the whole body (i.e. `@Body("a")`)
3. Have this interface extend any other interface

Separately using `@Body("a")` or inheritance works fine, but bringing them together creates the described outcome.

### Environment Information

```kotlin
plugins {
    id("io.micronaut.application") version "4.5.3"
    id("io.micronaut.aot") version "4.5.3"
    // ...
}

dependencies {
    annotationProcessor("io.micronaut.openapi:micronaut-openapi:6.15.0")
    compileOnly("io.micronaut.openapi:micronaut-openapi-annotations:6.15.0")
    // ...
}
```

### Example Application

_No response_

### Version

4.8.2

altro3:
Fixed #2136

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
@@ -264,9 +264,18 @@ public abstract class AbstractOpenApiEndpointVisitor extends AbstractOpenApiVisi
 
             if (CollectionUtils.isNotEmpty(superTypes)) {
                 ContextUtils.put(MICRONAUT_INTERNAL_IS_PROCESS_PARENT_CLASS, true, context);
-                List<MethodElement> methods = element.getEnclosedElements(ElementQuery.ALL_METHODS);
-                for (MethodElement method : methods) {
-                    visitMethod(method, context);
+                var hasSuperMethods = false;
+                for (var superType : superTypes) {
+                    if (CollectionUtils.isNotEmpty(superType.getMethods())) {
+                        hasSuperMethods = true;
+                        break;
+                    }
+                }
+                if (hasSuperMethods) {
+                    List<MethodElement> methods = element.getEnclosedElements(ElementQuery.ALL_METHODS);
+                    for (MethodElement method : methods) {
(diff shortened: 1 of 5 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
```diff
@@ -264,9 +264,18 @@ public void visitClass(ClassElement element, VisitorContext context) {
 
             if (CollectionUtils.isNotEmpty(superTypes)) {
                 ContextUtils.put(MICRONAUT_INTERNAL_IS_PROCESS_PARENT_CLASS, true, context);
-                List<MethodElement> methods = element.getEnclosedElements(ElementQuery.ALL_METHODS);
```
graemerocher, line 267: you may want to limit this to non static methods

graemerocher (approved):

openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiEndpointVisitor.java
altro3: great! Added `onlyInstance()`

PR #2167, opened 2025-05-31
Add ability to generate EnumConverterConfig - converters for endpoint parameters

altro3:


altro3:
Last feature for http clients / controllers. Need to describe custom type converters, if you use custom values / types for enums. Thi usefull, when your enum values with non String type or values not equals with enum constants names. Foe example like this:
```java
@Serdeable
@Generated("io.micronaut.openapi.generator.JavaMicronautClientCodegen")
public enum ShortPrimitiveEnum {

    @JsonProperty("1")
    NUMBER_1((short) 1),
    @JsonProperty("2")
    NUMBER_2((short) 2),
    @JsonProperty("3")
    NUMBER_3((short) 3),
    @JsonProperty("4")
    NUMBER_4((short) 4),
    @JsonProperty("5")
    NUMBER_5((short) 5),
    ;

    public static final Map<Short, ShortPrimitiveEnum> VALUE_MAPPING = Map.copyOf(Arrays.stream(values())
        .collect(Collectors.toMap(v -> v.value, Function.identity())));

    private final short value;

    ShortPrimitiveEnum(short value) {
        this.value = value;
    }
}
```

graemerocher (approved):

PR #2183, opened 2025-06-10
Improve classes output directrory and project directory detection

### LINKED ISSUE #2180: Non-Deterministic builds
When building using micronaut openapi generator, the final jar is not deterministic.
I ran `diff <(hexdump -C new-jar ) <(hexdump -C old-jar)`
And this was the result:
```
< 00174210  14 00 00 00 64 75 6d 6d  79 31 30 33 30 36 38 37  |....dummy1030687|
< 00174220  31 30 30 32 37 37 36 36  03 00 50 4b 03 04 14 00  |10027766..PK....|
---
> 00174210  14 00 00 00 64 75 6d 6d  79 31 30 31 33 36 38 30  |....dummy1013680|
> 00174220  34 32 38 36 36 35 36 35  03 00 50 4b 03 04 14 00  |42866565..PK....|
98464,98465c98464,98465
< 00180a10  6d 6d 79 31 30 33 30 38  32 30 37 34 35 36 39 32  |mmy1030820745692|
< 00180a20  34 36 03 00 50 4b 03 04  14 00 00 08 08 00 00 00  |46..PK..........|
---
> 00180a10  6d 6d 79 31 30 31 33 38  32 37 36 35 33 32 31 31  |mmy1013827653211|
> 00180a20  34 34 03 00 50 4b 03 04  14 00 00 08 08 00 00 00  |44..PK..........|
137975,137976c137975,137976
< 0021af80  00 a4 81 f6 41 17 00 64  75 6d 6d 79 31 30 33 30  |....A..dummy1030|
< 0021af90  36 38 37 31 30 30 32 37  37 36 36 50 4b 01 02 14  |68710027766PK...|
---
> 0021af80  00 a4 81 f6 41 17 00 64  75 6d 6d 79 31 30 31 33  |....A..dummy1013|
> 0021af90  36 38 30 34 32 38 36 36  35 36 35 50 4b 01 02 14  |68042866565PK...|
141023,141024c141023,141024
< 00226e00  73 77 61 67 67 65 72 2f  64 75 6d 6d 79 31 30 33  |swagger/dummy103|
< 00226e10  30 38 32 30 37 34 35 36  39 32 34 36 50 4b 01 02  |082074569246PK..|
---
> 00226e00  73 77 61 67 67 65 72 2f  64 75 6d 6d 79 31 30 31  |swagger/dummy101|
> 00226e10  33 38 32 37 36 35 33 32  31 31 34 34 50 4b 01 02  |382765321144PK..|
```
I'm guessing this is related to these code snippets:
https://github.com/micronaut-projects/micronaut-openapi/blob/bd4e9f0da5774af73310e1f2d4910eef2d2f3cce/openapi/src/main/java/io/micronaut/openapi/visitor/OpenApiApplicationVisitor.java#L890

https://github.com/micronaut-projects/micronaut-openapi/blob/bd4e9f0da5774af73310e1f2d4910eef2d2f3cce/openapi/src/main/java/io/micronaut/openapi/visitor/ContextUtils.java#L81

Not sure the whats the exact use-case for this, but is it possible to ues `java.util.Random` instead of `System.nanoTime` since Random is deterministic?

altro3:
Minimize creation of dummy files (now will be created only single dummy file).
Also now if user set `micronaut.openapi.project.dir` for KSP we can work without dummy files creation.

Fixed #2180


altro3:
ok, here is the final solution:

1. Dynamic postfixes for dummy files: yes, they are not required (at least I did not see any problems in their absence). I remember that I added this because one of my processors created this file and when I tried to call the getProjectDir method, I got an error because of it. Well, let's try to remove the postfix, we'll see what happens.

2. About deleting files: unfortunately, KSP does not provide any tools to delete the generated file. @dstepanov this is about your suggestion. Therefore, it is better not to create dummy files, because you will not be able to delete them

3. Limitations of KSP: unfortunately, KSP does not provide any way to find out the current directory of the project. The only way is to create that dummy file, which then cannot be deleted.

So what I did:

1. In this module, I just removed the postfix - I still don't believe that it somehow affects the cache, because the dummy file is created once for the entire process, the path to the project is calculated and this calculated value is saved in the context. The dummy file is not created anymore. Sorry, but I don't see where the caching problem could be here. And the gradle did not have problems with the cache if you ran the build several times. Looks like a far-fetched problem.

2. The main changes are here: https://github.com/micronaut-projects/micronaut-openapi/pull/2183

3. Now I try to calculate the path to the project directory and the output path whenever possible, so as not to create dummy files again later.

4. The main improvements concern KSP: now I can only create a single dummy file to calculate the project directory. All other paths are calculated automatically.

5. In the documentation, I added that for KSP it is highly desirable to specify the micronaut.openapi.project.dir property - in this case, dummy files are not created at all and the build will be clean.

I think this is the maximum that can be done in the difficult conditions that KSP provides.

sdelamo:
@altro3 can you merge 6.17.x into this branch and ensure the build passes?

docs-examples/example-java/build.gradle
```diff
@@ -28,7 +27,6 @@ compileTestJava.options.compilerArgs += '-parameters'
 test {
     jvmArgs '-Duser.country=US'
     jvmArgs '-Duser.language=en'
-    useJUnitPlatform()
```
dstepanov, line 31: BTW pls always try if tests are still running when you remove `useJUnitPlatform`. You did this in the sourcegen and tests were not picked up anymore

altro3: Actually, I didn't want to delete this line. BUT! Since junit 5.12 the tests just stopped working! Look at my changes - this is all a forced measure. But I checked - the tests run. As I understand it, now you need to explicitly register junit-platform-launcher in the dependencies, because it seems that gradle can no longer do this correctly on its own

Look to this: 

https://docs.gradle.org/8.12/userguide/upgrading_version_8.html#test_framework_implementation_dependencies

https://youtrack.jetbrains.com/issue/IDEA-370750/Idea-is-incompatible-with-JUnit-5.12

https://dev.to/be-hase/important-notes-on-junit-5120-in-gradle-13fj

PR #2235, opened 2025-06-21
Fixed Enum Case Sensitivity with non string values

altro3:
Fixed #2233

PR #2234, opened 2025-06-21
fix: fix kotlin-generator hideGenerationTimestamp produces invalid code

KDanisme:
Previous code doesn't compile with:

```
e: file:///tmp/test15651700443780886616/src/main/kotlin/org/openapitools/api/RequestBodyApi.kt:24:20 Assigning single elements to varargs in named form is forbidden
e: file:///tmp/test15651700443780886616/src/main/kotlin/org/openapitools/api/RequestBodyApi.kt:24:20 Type mismatch: inferred type is String but Array<out String> was expected
```

PR #2231, opened 2025-06-20
fix: fix kotlin client generator bytebuffer generics

KDanisme:
Fixes invalid java syntax in kotlin generator
`ByteBuffer<?>` -> `ByteBuffer<*>`

this is a leftover from the java generator

PR #2236, opened 2025-06-21
Add kotlin compilation testing for kotlin-client-generator

KDanisme:
So far I have found 3 Bugs using this method so I think it works quite well.

This PR Depends on:
#2231 
#2234 

I can comment out the non-working assertions for these tests if we don't want to wate for those PRs 

I have also commented out an assertion which should also be solved by #2235 

I have added these assertions for the client for now, Ill add them to the server if things go smoothly.
Java files also work with the compiler so I could add these tests to all the generators.

### DIFF
--- a/openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
+++ b/openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
@@ -1,13 +1,20 @@
 package io.micronaut.openapi.generator;
 
+import com.tschuchort.compiletesting.KotlinCompilation;
+import com.tschuchort.compiletesting.SourceFile;
+import org.apache.commons.io.FileUtils;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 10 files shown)


openapi-generator/src/test/resources/3_0/extra-annotations.yml
```diff
@@ -37,7 +37,7 @@ components:
     Book:
       type: object
       x-class-extra-annotation:
-        - "@Serializable"
+        - "@io.micronaut.serde.annotation.Serdeable.Serializable"
```
altro3, line 40: why you chaged this line?

KDanisme: Because when you add an extra annotation, the generator doesn't import it by default, meaning the previous code did not compile. 
I also modified the tests themselfs because they were testing uncompialable code

altro3:
@KDanisme gj! I think, need to add this compilation test for all tests - java client, java server and kotlin server

altro3:
@KDanisme hm... need to fix tests

KDanisme:
> @KDanisme hm... need to fix tests

Sorry must have only ran the kotlin client tests, maybe I broke something... checking.

KDanisme:
> @KDanisme gj! I think, need to add this compilation test for all tests - java client, java server and kotlin server

do you want this in this PR?
a lot of new issues might pop up so i thought id merge this first so you can debug and verify them easily

altro3:
@KDanisme Hm.. let's change base branch to 6.17.x also. And recreate PR, please. I think, we can add it to 6.17.1 release, because this is just improvement for tests. After it I can merge it and you can add this compile tests for java client/server and kotlin server and if we found some problems - wec fix it in 6.17.1. Don't need to wait next major release 6.18.0

KDanisme:
I changed the base without recreating,  and reset rebased on top of 6.17 let me know if its fine

altro3 (approved):

altro3:
> I changed the base without recreating, and reset rebased on top of 6.17 let me know if its fine

Yes, everything is fine now. But I lost my write acces :-( Don't know what happened

openapi-generator/src/test/java/io/micronaut/openapi/generator/AbstractMicronautCodegenTest.java
```diff
@@ -1,13 +1,20 @@
...
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
+import static org.junit.Assert.assertEquals;
```
altro3: Wrong import, need to use junit jupiter

graemerocher (approved):

PR #2244, opened 2025-06-24
Java generator fixes

altro3:
This is continue for https://github.com/micronaut-projects/micronaut-openapi/pull/2236.

Now we trying to compile generated files in every test.
Found and fixed some java generator problems


altro3:
@KDanisme thanks! Very usefull improvement!

KDanisme:
> @KDanisme thanks! Very usefull improvement!

I created a branch yesterday to do this as well :]
One issue I noticed in the compiler tests, is that the Junit reports don't have the compiler errors in them, we might want to add a message in the compilation result assertion to view the actuall compilation error. just a thought.

altro3:
It doesn't matter anymore. It's enough that the test fails if uncompiled code is generated.

graemerocher (approved):

PR #2269, opened 2025-07-06
Add property for Jackson include mode

### LINKED ISSUE #2261: Feature Request: Mark all schema properties as required
### Feature description

I have some classes/interfaces that represent API response types. I am configuring my API with Jackson's `JsonInclude.Include.ALWAYS` setting, so the fields will always be present in the response. I would rather not apply an `@Schema(requiredMode = Schema.RequiredMode.REQUIRED)` to every field, or list them all out in a class-level `@Schema(requiredProperties=...)` annotation at the class level.

I looked at the code and it seems that could probably do this in `SchemaDefinitionUtils.processSchemaAnn()`. I have a few ideas for how to solve this, but I'm not sure which would be the most proper approach:
1. Introduce a new annotation in micronaut-openapi that marks every property in the class/interface as required?
2. Introduce an application-level setting that allows changing this globally?
3. Interpret a special value of `@Schema(requiredProperties={"*"})` to mark all properties as required?

xpdota:
Closes #2261 . As suggested in the comments there, I implemented this as a system property.

This adds the `micronaut.openapi.property.include` setting which chooses an item from Jackson's `JsonInclude.Include` enum. If it is set to `ALWAYS`, then it will mark all properties as required in the schema. Currently, none of the other enum choices have any impact, as they still result in the field not necessarily being included.

I'm not sure if I entirely like the implication that this would also mark things as required on *request* schema types, since the intention is to mark everything as required on *response* schemas, which is what you would expect when using `Include.ALWAYS`. However, `ALWAYS` does not imply that all fields would be required on a request. Very open to feedback as to what the most proper way would be.

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/ConfigUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/ConfigUtils.java
@@ -1195,31 +1195,16 @@ public final class ConfigUtils {
         };
     }
 
-    public static JsonInclude.Include getInclude(VisitorContext context) {
+    public static boolean isIncludeAll(VisitorContext context) {
         var value = getConfigProperty(MICRONAUT_OPENAPI_PROPERTY_INCLUDE, context);
         if (value == null) {
-            return null;
+            return false;
         }
-        return toJacksonInclude(value.toUpperCase(Locale.ENGLISH), context);
-    }
-
-    private static JsonInclude.Include toJacksonInclude(String include, VisitorContext context) {
-        if (include == null) {
-            return null;
+        // ALWAYS is the only value that has any impact on API spec generation.
+        if (value.toUpperCase(Locale.ENGLISH).equals("ALWAYS")) {
+            return true;
         }
-        return switch (include.toUpperCase(Locale.ENGLISH)) {
-            case "ALWAYS" -> JsonInclude.Include.ALWAYS;
-            case "NON_NULL" -> JsonInclude.Include.NON_NULL;
-            case "NON_ABSENT" -> JsonInclude.Include.NON_ABSENT;
-            case "NON_EMPTY" -> JsonInclude.Include.NON_EMPTY;
-            case "NON_DEFAULT" -> JsonInclude.Include.NON_DEFAULT;
-            case "CUSTOM" -> JsonInclude.Include.CUSTOM;
-            case "USE_DEFAULTS" -> JsonInclude.Include.USE_DEFAULTS;
-            default -> {
-                warn("Unknown include value: " + include, context);
-                yield null;
-            }
-        };
+        return false;
     }
 
     public static String getConfigProperty(String key, VisitorContext context) {
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/ConfigUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/ConfigUtils.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.openapi.visitor;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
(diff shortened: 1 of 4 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/ConfigUtils.java
```diff
@@ -1193,6 +1195,33 @@ private static PropertyNamingStrategies.NamingBase toJacksonStrategy(String nami
...
+    }
+
+    private static JsonInclude.Include toJacksonInclude(String include, VisitorContext context) {
+        if (include == null) {
+            return null;
+        }
+        return switch (include.toUpperCase(Locale.ENGLISH)) {
+            case "ALWAYS" -> JsonInclude.Include.ALWAYS;
+            case "NON_NULL" -> JsonInclude.Include.NON_NULL;
+            case "NON_ABSENT" -> JsonInclude.Include.NON_ABSENT;
```
altro3: Don't neeed to add unsupported values. micronaut-openapi can correctly interpret only `NON_NULL` and `ALWAYS`

openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
```diff
@@ -1722,6 +1723,11 @@ public static void processSchemaProperty(VisitorContext context, TypedElement el
                 required = true;
             }
 
+            // If inclusion is ALWAYS, then all properties are required
+            if (!required && ConfigUtils.getInclude(context) == JsonInclude.Include.ALWAYS) {
```
altro3: I don't think, that is enought... There are clear priorities in code processing (from least to most):

1. Interpretation of code without annotations
2. Annotations by Microsoft and Jackson
3. Annotations by Swagger

That is, if the value was calculated as mandatory in the first and second stages, but the Swagger annotation states that the property is not mandatory - the value from the Swagger annotation should be taken.

The same with Jackson annotations - you can override the default behavior for a specific class or property using the JsonInclude annotation at the class or property level.

I say this because more tests clearly need to be written and the feature is not as simple as it might seem at first glance

xpdota: Yeah, I'm realizing that it isn't as simple as I had thought. I've updated it to account for a Swagger annotation reverting the property to non-required, but I'm not sure about \#2 - I understand that a property- or class-level Jackson or Micronaut annotations (I assume "Microsoft" was an autocorrect?). I see that `JsonProperty.required` is checked in `isNotNullable`, but is there logic elsewhere that I would need to take into account?

I've pushed a commit with a couple changes so far.

altro3:
@xpdota I review it later. Haven't time right now. But I can say: need add full support for JsonInclude annotation. Now you can change default behaviour for required field in schema. But need to add ability to change default behaviour on class and property level - with annotaion JsonInclude

xpdota:
> @xpdota I review it later. Haven't time right now. But I can say: need add full support for JsonInclude annotation. Now you can change default behaviour for required field in schema. But need to add ability to change default behaviour on class and property level - with annotaion JsonInclude

The more I think about it, the more problems there are with my approach. This logic works fine for a response type, where "required" means "this field will always be present, even if it's null/empty". But that isn't true for a request type - setting up the serializer to include all fields (whether globally, per-class, or per-field) on a particular model doesn't mean that those are necessarily required for deserialization. That is, this change would make it *less* correct for request bodies.

As far as I can tell, the OpenAPI spec doesn't really support this concept - a field can't be required in one direction, but optional in the other. Thus, the only way to make this work would be to have corresponding "request" and "response" versions of any object that appears in both a request and a response.

I'm thinking it might make more sense to just fix this on the generator side? In my case, it's a Typescript consumer, so I can make all properties required by wrapping the type with `Required<T>`.

altro3:
@xpdota I improved your PR here https://github.com/micronaut-projects/micronaut-openapi/pull/2337 (dont kow how to chnage cuurent PR)

I've added fulls support for `JsonInclude` annotation and also add some more tests for it. Continue discussion here: https://github.com/micronaut-projects/micronaut-openapi/pull/2337

PR #2295, opened 2025-07-27
Add ability to exclude classes and packages from processing

altro3:


openapi-annotations/src/main/java/io/micronaut/openapi/annotation/OpenAPIExclude.java
```diff
@@ -0,0 +1,63 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.RetentionPolicy.SOURCE;
+
+/**
+ * The annotation can be used to include additional io.micronaut.http.annotation.Controller or
```
graemerocher: javadoc description seems incorrect

graemerocher (changes requested):

openapi-annotations/src/main/java/io/micronaut/openapi/annotation/OpenAPIExclude.java
altro3: fixed

graemerocher (approved):

PR #2296, opened 2025-07-29
Introduce new behavior with JsonUnwrapped and Schema(allOf) annotations

uvoigt:
@altro3 

add another PR to this repository


altro3:
Need test for classes with same property and different types. Also question: did you check how it works with real controller? I don't think this will work correct

uvoigt:
Code is aware of duplicates in properties now. The first occurrence of a property within a type n the `allOf` list has precedence over following occurrences.
Yes, I tested it with the real scenario where I was confronted with the lack of support for `@JsonUnwrapped` and `@Schema(allOf)`.

altro3 (approved):

PR #2307, opened 2025-08-02
Remove micronaut-inject dependency from some modules

### LINKED ISSUE #2297: micronaut-openapi-annotations depends on micronaut-core-bom
We upgraded to 6.16 to work around an issue and discovered that this force-upgraded us to core 4.8 (we are on 4.6).

Please remove this dependency!

altro3:
Fixed gradle scripts for correct testing.
Fixed documentation for using with SpringBoot

Fixed #2297

### DIFF
--- a/openapi-annotations/build.gradle
+++ b/openapi-annotations/build.gradle
@@ -1,7 +1,14 @@
 plugins {
-    id 'io.micronaut.build.internal.openapi-module'
+    id 'io.micronaut.build.internal.openapi-simple-module'
 }
 
 dependencies {
     api(libs.managed.swagger.annotations)
+}
+
+configurations.configureEach {
+
+    exclude group: "io.micronaut", module: "micronaut-inject-java"
+    exclude group: "io.micronaut", module: "micronaut-inject"
+    exclude group: "io.micronaut", module: "micronaut-core-bom"
 }
--- a/build-logic/src/main/groovy/io.micronaut.build.internal.openapi-base.gradle
+++ b/build-logic/src/main/groovy/io.micronaut.build.internal.openapi-base.gradle
@@ -3,6 +3,7 @@ repositories {
 }
 
 tasks.withType(Test).configureEach {
+    useJUnitPlatform()
     testLogging {
         exceptionFormat = 'full'
     }
--- a/build-logic/src/main/groovy/io.micronaut.build.internal.openapi-java-generator-test-suite.gradle
+++ b/build-logic/src/main/groovy/io.micronaut.build.internal.openapi-java-generator-test-suite.gradle
@@ -56,3 +56,10 @@ dependencies {
         implementation("io.micronaut:micronaut-http-client:${libs.versions.micronaut.asProvider().get()}")
     }
 }
+
+tasks.withType(Test).configureEach {
+    useJUnitPlatform()
+    testLogging {
+        exceptionFormat = 'full'
+    }
+}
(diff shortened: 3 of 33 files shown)


altro3:
@graemerocher @sdelamo @dstepanov @yawkat So, this is the only quick and working solution I could come up with.

So, in general, everything is simple - we remove dependencies on micronaut-inject in those modules where we don't need it.

But I had to add a "cheat" - I had to duplicate the annotations `io.micronaut.context.annotation.Aliases` and `io.micronaut.context.annotation.AliasFor` to remove the dependency on `micronaut-inject` in `micronaut-openapi-annotations`

I think this is the only right decision in these circumstances.

graemerocher:
we can't duplicate classes like this. It is a hack that will come back to bite later.

altro3:
How do you suggest to solve this problem?

graemerocher:
The underlying issue report imo is not valid. It is unlikely that this module will work with an earlier version. The user should upgrade.

altro3:
maybe, just remove these annotations and fix processing of annotations which used `@AliasFor`?

altro3:
If I remove the use of this annotation, it will solve all the problems. I suggest doing so.

altro3:
@graemerocher In general, I think it is more correct to apply these changes - after them, each model will depend only on those libraries that it really needs. I don't see any problems with this solution.

Yes, this will not solve the problem with what was described in the ticket 100%, but my goal was to finally remove all transitive dependencies that are not actually required in some modules.

What you wrote about the problem in the ticket: I completely agree that this can lead to incorrect operation of the application, but as I wrote above, the goal of this PR is to correct the dependencies.

Yes, with Spring Boot. unfortunately, we will not be able to solve the problem, because for this we need to separate the inject module into a module with annotations and a module with logic. But this is not so critical.

In general, I think that this PR will finally put the module dependencies in order and they will be transparent, and most importantly, there will be nothing unnecessary in the modules

dstepanov:
Just put it as a compile only.

openapi-annotations/build.gradle
```diff
@@ -1,7 +1,14 @@
...
+    id 'io.micronaut.build.internal.openapi-simple-module'
 }
 
 dependencies {
     api(libs.managed.swagger.annotations)
 }
+
+configurations.configureEach {
+
+    exclude group: "io.micronaut", module: "micronaut-inject-java"
```
dstepanov, line 11: FYI I belive this syntax is deprecated in Gradle

altro3: really? hm... what is correct syntax? I didn't find any information about it

altro3: I mean, I haven't seen any messages about this syntax being deprecated. Also, I haven't seen any information about it on the internet. And the official docs say that dependency exploitation should be done like this

dstepanov: Not sure where I saw it, I think the prefered syntax is the full string one: "io.micronaut:micronaut-inject-java"

altro3: @dstepanov I didn't find any information about deprecation of this syntax. Also I didn't find how to do it another way.

Also I moved all scripts to kotlin DSL here: https://github.com/micronaut-projects/micronaut-openapi/pull/2321

And with Kotlin DSL I found only this syntax.

graemerocher (approved):

PR #2310, opened 2025-08-04
[openapi] Update common files for branch 6.18.x

micronaut-build:
Update common files


altro3:
@graemerocher @dstepanov Why you updated gradle to 9.0 ?

The problem is that officially micronaut supports working with kotlin 1.9.25. Accordingly, it supports working with kapt version 1.9.25.

But here's the problem: kapt 1.9.25 uses deprecated gradle api and is incompatible with gradle 9.0.

I suggest rolling back the gradle version to 8.14.3

dstepanov:
It was merged automatically

altro3:
need to revert it... we can't use gradle 9.0 with micronaut 4.x

altro3:
@graemerocher also this, pls. Same problem

altro3 (approved):

PR #2311, opened 2025-08-04
Added annotation `@OpenAPIRequest` as alternative to `@io.micronaut.http.annotation.RequestBean` annotation

altro3:
This annotation needed, to remove `micronaut-http` from Spring Boot applications

### DIFF
--- a/openapi-annotations/src/main/java/io/micronaut/openapi/annotation/OpenAPIRequest.java
+++ b/openapi-annotations/src/main/java/io/micronaut/openapi/annotation/OpenAPIRequest.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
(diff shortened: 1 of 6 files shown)


altro3:
@graemerocher You asked me why I use the RequestBean annotation in the Spring Boot controller - I told you that Spring Boot does not have such annotation and it processes parameters correctly without any additional annotations.

The solution with RequestBean annotations is completely correct and works, and micronaut-http must currently be in the compilation dependencies for a Spring Boot application.

But I want to completely get rid of the dependency on micronaut libraries in Spring Boot applications (that is, leave them at the annotation processor level).

Consider this a continuation of https://github.com/micronaut-projects/micronaut-openapi/pull/2307

graemerocher (approved):

PR #2320, opened 2025-08-17
Improved spring boot docs and added test suites for Spring + KAPT and Spring + KSP

altro3:


altro3:
Merge it after #2311 and #2307!

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
```diff
@@ -0,0 +1,73 @@
+Annotation api:openapi.annotation.OpenAPIRequest[] can be used as alternative to *io.micronaut.http.annotation.RequestBean* to mark parameter as wrapped request details. It useful, when you haven't micronaut-http in dependencies. For example, when you use micronaut-openapi with spring-boot application.
```
graemerocher, line 1: describe what dependency is needed on the path for this. It can't be the `openapi` module since that will pull in all the processing dependencies

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+.Example of using `@OpenAPIRequest` annotation with a Spring Boot application
+[source,java]
+----
+@RestController
+@RequestMapping("/my-api")
+public class MyController {
```
graemerocher: this should be an include pulled from sources with a kotlin example

graemerocher (changes requested):

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
altro3: Could you be more specific about what needs to be fixed? I didn't understand what needs to be added.

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
altro3: How can I make it so that there are examples from Kotlin and Java at the same time?

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
graemerocher: In order for the user to reference `@OpenAPIRequest` they would need a module on the class path. I assume it is this module https://github.com/micronaut-projects/micronaut-openapi/tree/6.18.x/openapi-annotations

this is not mentioned anywhere

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
graemerocher: Using the `snippet:` macro. See examples in the documentation of our other projects.

altro3: fixed

src/main/docs/guide/micronautOpenApiAnnotations/openApiRequest.adoc
altro3: I looked at the documentation again: information regarding the mcironaut-openapi-annotations dependency is here:

https://github.com/micronaut-projects/micronaut-openapi/blob/6.18.x/src/main/docs/guide/micronautOpenApiAnnotations.adoc

altro3:
@graemerocher Also I've added another one change:

{99B13458-2904-4635-98B3-11E58F969B99}

These 3 libraries are necessary for correct work with Spring Boot. These 3 libraries are in any case in the classpath of any application on Micronaut. But if we add them as dependencies to micronaut-openapi-annotations - this will simplify the connection to the application on Spring Boot, i.e. for correct work it will be enough to add only:
```kotlin
compileOnly("io.micronaut.openapi:micronaut-openapi-annotations")
```

graemerocher:
you cannot use `*` with excludes on dependencies when building a library. Only explicit excludes for each transitive will work.

altro3:
fixed

{EA5CD630-896F-449D-9544-D15E65C8564A}

graemerocher (approved):

PR #2330, opened 2025-08-24
Added ability to change javadoc conversion mode: HTML_TO_MD, MD_TO_HTML, PLAIN.

altro3:
Added more extensions for flexmark to more correct conversion.


altro3:
@graemerocher @melix Could you elp with this error? 

{38428CA6-987A-4BD3-918B-2055200371E8}


I understand what the problem is, but I don't think it's a bug. This library is not used anywhere except this module. Is there any way to add this fix to the exception?

And also: I just set more correct name for library version:

{3899CBDF-BDE4-4BC3-86EE-85B7925CDD71}

melix:
You get this error because you removed a managed dependency, that is to say a dependency which may be used by clients that use the Micronaut version catalog. This would break their builds. As such it's a breaking change. So when you say:

> This library is not used anywhere except this module.

Either the library should never have been managed in the first place and should have remained an internal implementation detail, or it's correctly managed and you have a breaking change.

altro3:
@melix Ok, I reterned old version and library, but marked them as deprecated

graemerocher (approved):

PR #2343, opened 2025-09-02
Added ability to set patterns (and exclude patterns) for additional files.

### LINKED ISSUE #2259: "micronaut.openapi.additional.files" fails to merge the .yml files when value is prefixed with "classpath:"
### Expected Behavior

While trying to merge OpenAPI YAML files, I set `micronaut.openapi.additional.files=classpath:META-INF/swagger` 

According to the docs, the `classpath:` prefix should resolve resources inside JARs and all the .yml files in `META-INF/swagger` should be merged into one.

### Actual Behaviour

However, it generates a WARNING `([WARNING] Fail to load META-INF/swagger`) when `classpath:` is used. The same value works when:

- I replace classpath: with project and use a folder in my project:
`project:src/main/resources/META-INF/swagger`

- I point to an explicit path:
`target/classes/META-INF/swagger` without any prefix

But the prefix "classpath:" is really necessary to my project.

### Steps To Reproduce

1. Open the example project;
2. Run `mvn clean compile`
3. See the log `[WARNING] Fail to load META-INF/swagger`

### Environment Information

- JAVA 21

### Example Application

https://github.com/BiancaPuertaRocha/test-merge-swagger

### Version

4.9.0

altro3:
Fixed #2259


altro3:
@graemerocher @sdelamo Something is wrong with github action:

{74163261-898E-46A0-A147-EC429A62CDF0}

graemerocher (approved):

PR #2378, opened 2025-09-28
Added Serdeable annotation to oneOf interface

### LINKED ISSUE #2377: oneof models are missing Serdeable annotation
### Expected Behavior

when generating oneof interfaces they should be Serdeable just like all other models.

### Actual Behaviour

when generating oneof interfaces the resulting interfaces looks like this:
```
@Generated("io.micronaut.openapi.generator.KotlinMicronautClientCodegen")
interface AccessLevel
```
which makes http calls using it error with:
`Error decoding HTTP response body: No bean introspection available for type [interface <redacted>.model.AccessLevel]. Ensure the class is annotated with io.micronaut.core.annotation.Introspected`

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

6.18.1

altro3:
Fixed #2377

### DIFF
--- a/openapi-generator/src/main/resources/templates/java-micronaut/common/model/enum.mustache
+++ b/openapi-generator/src/main/resources/templates/java-micronaut/common/model/enum.mustache
@@ -19,6 +19,9 @@
 {{#micronaut_serde_jackson}}
 @Serdeable
 {{/micronaut_serde_jackson}}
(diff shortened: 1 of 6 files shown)


sdelamo:
@altro3 fix the conflict please

PR #2383, opened 2025-09-30
Fix test after upgrade micronaut-data

altro3:


altro3:
@graemerocher could you merge this PR first? And after it I'll rebase all other PR's

PS which branch must be? 6.18.x or 6.19.x ?

PR #2399, opened 2025-10-08
Bump core + fixes

dstepanov:


dstepanov:
@altro3 Now it's possible to use `element.getDocumentation(true)` to get parsed Javadoc for `getReturnType`, method, fields, parameters. Possible some of the logic can be refactored

gradle/libs.versions.toml
```diff
@@ -26,7 +26,7 @@ guava = "33.4.8-jre"
 commonmark = "0.26.0"
 kotlin-compile-testing = "1.6.0"
 
-micronaut = "4.10.2"
+micronaut = "4.9.11"
```
graemerocher, line 29: why the downgrade?

altro3:
@dstepanov why you closed it?

dstepanov:
I will try to fix core to have the previous behaviour

PR #2416, opened 2025-10-14
don’t define freemarker version, use views version

sdelamo:
I don't want to have freemarker version defined in multiple places


altro3 (approved):

altro3:
@sdelamo Please, rebase it to 6.19.x

Issue #2564, opened 2026-01-16
Micronaut Launch: The sample project fails when adding openapi


viniciusxyz:
### Expected Behavior

When generating a project with OpenAPI, I expect to be able to make requests without problems.

### Actual Behaviour

When I try to make a request, the following exception is thrown:

```sh
java.lang.NoClassDefFoundError: io/micronaut/core/async/publisher/Publishers
	at io.micronaut.http.server.ResponseLifecycle.encodeHttpResponse(ResponseLifecycle.java:137)
	at io.micronaut.http.server.ResponseLifecycle.encodeHttpResponseSafe(ResponseLifecycle.java:108)
	at io.micronaut.http.server.netty.RoutingInBoundHandler.writeResponse(RoutingInBoundHandler.java:217)
	at io.micronaut.http.server.netty.NettyRequestLifecycle.handleNormal(NettyRequestLifecycle.java:105)
	at io.micronaut.http.server.netty.RoutingInBoundHandler.accept(RoutingInBoundHandler.java:204)
	at io.micronaut.http.server.netty.handler.PipeliningServerHandler$MessageInboundHandler.read(PipeliningServerHandler.java:467)
	at io.micronaut.http.server.netty.handler.PipeliningServerHandler.channelRead(PipeliningServerHandler.java:241)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:356)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelRead(CombinedChannelDuplexHandler.java:434)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:361)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:325)
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:249)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:354)
	at io.netty.handler.timeout.IdleStateHandler.channelRead(IdleStateHandler.java:293)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:354)
[...]

Issue #2506, opened 2025-12-05
Transitive exclusions break dependency resolution in Maven


kristofdho:
### Expected Behavior

all dependencies correctly listed in `mvn dependency:tree`, possibly linked to this: https://github.com/micronaut-projects/micronaut-openapi/issues/2297
I tracked it to I think the transitive dependency exclusions present in the latest versions of the openapi annotations module.

### Actual Behaviour

one example, `micronaut-context-propagation` is missing, leading to no class def exceptions

### Steps To Reproduce

1. https://micronaut.io/launch/?type=DEFAULT&name=demo&package=com.example&javaVersion=JDK_17&lang=JAVA&build=MAVEN&test=JUNIT&features=openapi&version=4.10.3
2. `mvn dependency:tree -Dverbose`
3. `micronaut-context-propagation` is missing
4. changing verion to `4.9.4` restores dependency resolution
5. removing the openapi dependency also restores dependency resolution

### Environment Information

JDK 21 or 25
OS Windows

### Example Application

https://micronaut.io/launch/?type=DEFAULT&name=demo&package=com.example&javaVersion=JDK_17&lang=JAVA&build=MAVEN&test=JUNIT&features=openapi&version=4.10.3

### Version

4.10.3

PR #2565, opened 2026-01-30
Remove exclude for micronaut-core-reactive

lissam1:
Closes: #2564

### DIFF
--- a/openapi/build.gradle.kts
+++ b/openapi/build.gradle.kts
@@ -11,11 +11,7 @@ dependencies {
 
     implementation(mn.micronaut.core)
     implementation(mn.micronaut.inject)
-    implementation(mn.micronaut.http) {
-        exclude(group = "io.micronaut", module = "micronaut-context")
-        exclude(group = "io.micronaut", module = "micronaut-context-propagation")
-        exclude(group = "io.projectreactor", module = "reactor-core")
-    }
+    implementation(mn.micronaut.http)
 
     api(projects.micronautOpenapiAnnotations)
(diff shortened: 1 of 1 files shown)


altro3 (approved):

altro3:
@lissam1 Just remove all excusions from build gradle script. I added this to remove unnecessary libraries from dependencies, but since this causes problems for maven, you can remove these exclusions.

{BC825884-5DF2-4946-B40C-26E0541CC663}

altro3:
But I'm not sure, why  it works fine with gradle, but with some problems with maven. Maybe need to check `micronaut-starter` prooject and fix it

lissam1:
> But I'm not sure, why it works fine with gradle, but with some problems with maven. Maybe need to check `micronaut-starter` prooject and fix it

i checked on starter only difference i spot is how dependencies being resolved in gradle we use bom from io.micronaut.application and in maven we use directly micronaut-parent, and also it may be of scopes as in maven we use provided

lissam1:
i test this change locally i still get the failure, i think we may need to update starter 
@altro3

altro3:
Conclusion: Don't use Maven :-)

altro3:
@lissam1 look to this, please: https://github.com/micronaut-projects/micronaut-openapi/issues/2506

Looks like the same problem

lissam1:
> @lissam1 look to this, please: #2506
> 
> Looks like the same problem

yes check same issue gradle has dependency as a transitive but in maven it get excluded, i check starter no sign for how it get excluded.

PR #2577, opened 2026-02-15
Update libs

altro3:


openapi-generator/build.gradle.kts
```diff
@@ -16,6 +16,7 @@ dependencies {
     api(mnLogging.slf4j.ext)
     api(mn.snakeyaml)
     api(mn.jackson.datatype.jsr310)
+    api(libs.rhino)
```
graemerocher: why do we need rhino?

altro3: Unfortunately, otherwise the build won't build, and SonaType complains that the library contains a vulnerability. We don't use this library, but it is included as a transitive dependency via openapi-generator , which references the old swagger-parser, which in turn references rhino.

If you have another solution to this problem, no problem. Let's find a different solution. But the fact remains: if we don't do this, the build will fail.

graemerocher: we should get a new transitive of the openapi-generator that fixes this vulnerability

altro3: I don't quite understand what you're talking about... That's exactly what I did, replacing the vulnerable version with a non-vulnerable one. All other versions are up to date, but Rhino still connects to the old version, and there's no way to fix it without manual intervention.

altro3: @graemerocher That's the reason, why  I did it: https://github.com/micronaut-projects/micronaut-openapi/actions/runs/22398072224/job/64836816791

изображение

sdelamo:
```suggestion
    api(libs.rhino) // TODO remove once `org.openapitools:openapi-generator` does not ship with a vulnerable version of rhino
```

sdelamo (approved):

Issue #2572, opened 2026-02-11
Post Requests results in Mono<HttpResponse<thing>> rather than Mono<thing>


Jonathan-Zollinger:
### Expected Behavior

creating a reactive client doesn't return any HttpResponse<> objects

### Actual Behaviour

SOME (not all) clients produce a `Mono<HttpResponse<thing>>` response

### Steps To Reproduce

all of this is available at [this commit](https://github.com/PeanutButter-Unicorn/z4j/tree/d59f7704564f12604f91964cc51a5094eaa5a2fa), you can just skip to step 3 

1. configure openapi in build file with these settings, using a reactive client and setting the generator to not generate httpResponses:
```kotlin
micronaut {
    runtime("netty")
    testRuntime("spock2")
    processing {
        incremental(true)
        annotations("lol.pbu.*")
    }
    openapi {
        version = "6.19.3"
        client(file("src/main/resources/z4j.yaml")) {
            apiPackageName.set("lol.pbu.z4j.client")
            modelPackageName.set("lol.pbu.z4j.model")
            useReactive.set(true)
            useAuth.set(false)
            lombok.set(true)
            clientId.set("zendesk")
            apiNameSuffix.set("Client")
            alwaysUseGenerateHttpResponse.set(false)
        }
    }
}
```
2. add post request to client file
```
...
  /api/v2/help_center/user_segments:
    post:
      operationId: CreateUserSegment
      tags:
        - User Segments
      summary: Create User Segment
      description: |
        (Allowed for Help Center managers)
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                user_segment:
                  $ref: '#/components/schemas/UserSegment'
      responses:
        "201":
          description: Created response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/UserSegmentResponse'
```
3. call build, `./gradlew clean build`
4. view compiled class:
```java
...
    /**
     * {@summary Create User Segment}
     * (Allowed for Help Center managers)
     *
     * @param createUserSegmentRequest (optional)
     *
     * @return Created response (status code 201)
     *         or Bad request response (status code 400)
     */
    @Post("/api/v2/help_center/user_segments")
    Mono<HttpResponse<@Valid UserSegmentResponse>> createUserSegment(
        @Body @Nullable @Valid CreateUserSegmentRequest createUserSegmentRequest
    );
```

### Environment Information

- OS: Fedora 43
[...]

PR #2579, opened 2026-02-16
Improved wrapping in HttpResponse. Added `@Status` annotation to generated controllers

### LINKED ISSUE #2563: 404 Response codes in kotlin client generator
### Feature description

When using the kotlin client generator, responses having a "404" in the spec responses do not make the return type of the request nullable.
Nor do they cause it to be come an HttpResponse using the `generateHttpResponseWhereRequired` flag, this flag only cares about 2xx responses.

This isn't an issue in java, but in kotlin these types that aren't null are presumed to be not null by the kotlin compiler, meaning it marks null checking code as redunant.

I propose either to make it so any extra response codes convert the response to an HttpRequest.
Or at least  make it nullable, but sometimes 404 responses do contain bodies, so that could be confusing.

altro3:
Fixed #2563

Ok, one PR for everything related to HttpResponse wrapping.

1. It now checks for a single 2xx response, and if so, the response is NOT automatically wrapped in the client and controller, and a Status annotation with the return code is added to the controller.

2. I added support for two extensions: `x-http-response-wrapper` and `x-not-null`.

3. Now return type will be with question mark (nullable), if response schema with `nullable: true` flag

You can now explicitly enable or disable the response wrapper in HttpResponse for each operation individually:

If `x-http-response-wrapper: true`, then the HttpResponse wrapper will definitely be added.

If `x-http-response-wrapper: false`, then the HttpResponse wrapper will definitely NOT be added.

If `x-http-response-wrapper` is missing, then the HttpResponse wrapper will be added depending on the global settings - generateHttpResponseAlways or generateHttpResponseWhereRequired.

`x-not-null` works the same way, but currently only works with the return value of operations and only in the Kotlin generator. This was added to allow manual adjustments to the generator if the return type should be nullable but not-nullable is generated. Basically, whether or not to add a question mark at the end of the type

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
@@ -81,6 +81,23 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static io.micronaut.openapi.generator.Extension.EXT_ANNOTATIONS_CLASS;
+import static io.micronaut.openapi.generator.Extension.EXT_ANNOTATIONS_FIELD;
+import static io.micronaut.openapi.generator.Extension.EXT_ANNOTATIONS_OPERATION;
+import static io.micronaut.openapi.generator.Extension.EXT_ANNOTATIONS_SETTER;
+import static io.micronaut.openapi.generator.Extension.EXT_DEPRECATED;
+import static io.micronaut.openapi.generator.Extension.EXT_DEPRECATED_MESSAGE;
+import static io.micronaut.openapi.generator.Extension.EXT_ENUM_DEPRECATED_MESSAGES;
+import static io.micronaut.openapi.generator.Extension.EXT_ENUM_DESCRIPTIONS;
+import static io.micronaut.openapi.generator.Extension.EXT_ENUM_VAR_NAMES;
+import static io.micronaut.openapi.generator.Extension.EXT_FORMAT;
+import static io.micronaut.openapi.generator.Extension.EXT_HTTP_RESPONSE_WRAPPER;
+import static io.micronaut.openapi.generator.Extension.EXT_TYPE;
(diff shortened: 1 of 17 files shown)


graemerocher:
can this be documented? Are there breaking changes to the defaults here?

altro3:
I wouldn't call these breaking changes... Rather, it's a more accurate understanding of when it's actually necessary to return an HttpResponse . Previously, with the generateHttpResponseWhereRequired flag, it was generated even if you only had one successful response code. That is, if at least two responses were described, for example, 404 and 200, it was generated, even though it shouldn't have been.

The status annotation was added for cases where you have a single successful response described, but with a non-standard code, such as 204.

The rest are new options for adjusting the generator using extensions, if, for example, you definitely need an HttpResponse but it isn't generated, or vice versa.

So I'd call this a bug fix and functionality improvement, but certainly not breaking changes.

Regarding documentation: unfortunately, no one has written documentation for the generator yet, and I don't have the time.

graemerocher:
so if the user updates the generator they may get different stubs generated for the server no? For example by returning an HttpResponse you have access to the headers, but in this case you do not

altro3:
In theory, yes, some methods that previously incorrectly returned an HttpResponse will now return the unwrapped schema class, as they were originally intended to do.

altro3:
Main problem was described here: https://github.com/micronaut-projects/micronaut-openapi/issues/2572

Also I fixed problem with nullable / not-nullable reponses for kotlin: https://github.com/micronaut-projects/micronaut-openapi/issues/2563

And also I've added 2 new extensions for help: `x-not-null` and `x-http-response-wrapper` to customize generated code

graemerocher:
this should be opt-in and we should avoid breaking existing applications

altro3:
@sdelamo wait with release, please, I'll fix this PR tomorrow

altro3:
@graemerocher @sdelamo Ok, added new option `generateHttpResponseOnlyForMultipleStatuses` to turn on new logic for automatical wrapping. By default all works as before

altro3:
@graemerocher @sdelamo could you merge it?

graemerocher (approved):

PR #2585, opened 2026-02-23
Issue 2584 model with additionalProperty create an invalid import

### LINKED ISSUE #2584: openapi-generator(kotlin): model with additionalProperty create an invalid import for java.util.HashMap
### Expected Behavior

The generated model should compile
There should be no import for `HashMap` because there is already an import for `java.util.*`

### Actual Behaviour

The generator produce an import for `org.openapitools.model.HashMap` which does not exists.

### Steps To Reproduce

```
components:
  schemas:
    ExampleDto:
      required:
        - field
      properties:
        field:
          description: 'Field'
          type: string
      type: object
      additionalProperties: true
```

### Environment Information

6.19.x branch

### Example Application

_No response_

### Version

4.10.8

etremblay:
Fix #2584


etremblay:
An alternative approach would to use kotlin `MutableMap` instead of `java.util.HashMap` as the base class for pojo with additional properties

altro3 (approved):

etremblay:
Thanks :slightly_smiling_face:

PR #2610, opened 2026-03-10
changes to groovy 5 instanceof verifications

sdelamo:


sdelamo:
@copilot can you search if there are changes in Apache Groovy 5 related to `instanceof` checks

PR #2645, opened 2026-03-27
Added support `enumUnknownDefaultCase` and `enumUnknownDefaultCaseEnumName` options for generator

### LINKED ISSUE #2642: enumUnknownDefaultCase does not update fromValue()
### Expected Behavior

 When enumUnknownDefaultCase is set to true via additionalProperties, the generator correctly adds the UNKNOWN_DEFAULT_OPEN_API constant to the enum, but the fromValue() method is not updated to return it for unknown values.

It still throws IllegalArgumentException, making the constant useless for its
  intended purpose.

 Expected behavior:
```java
  @JsonCreator
  public static Status fromValue(String value) {
      Status result = VALUE_MAPPING.get(value);
      if (result == null) {
          return UNKNOWN_DEFAULT_OPEN_API; // safe fallback for unknown values
      }
      return result;
  }
```

### Actual Behaviour

```java
  @JsonCreator
  public static Status fromValue(String value) {
      if (!VALUE_MAPPING.containsKey(value)) {
          throw new IllegalArgumentException("Unexpected value '" + value + "'"); // always throws
      }
      return VALUE_MAPPING.get(value);
  }
```

### Steps To Reproduce

1. Add an client for an api.yml that contains enum
2. set `additionalProperties.put("enumUnknownDefaultCase", "true")`in the build.gradle
3. Enum is generated with enum value `UNKNOWN_DEFAULT_OPEN_API("unknown_default_open_api");
`, but `fromValue` doesn't default to this value causing an unknown value to fail

### Environment Information

- Micronaut 4.10.8
- Plugin  io.micronaut.openapi 4.6.1 

### Example Application

https://github.com/leo-geoffroy-sonarsource/micronaut-openapi-enum-reproducer

### Version

4.10.8

altro3:
Fixed #2642

### DIFF
--- a/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
+++ b/openapi-generator/src/main/java/io/micronaut/openapi/generator/AbstractMicronautJavaCodegen.java
@@ -129,6 +129,7 @@ import static io.swagger.v3.parser.util.SchemaTypeUtil.FLOAT_FORMAT;
 import static io.swagger.v3.parser.util.SchemaTypeUtil.INTEGER64_FORMAT;
 import static io.swagger.v3.parser.util.SchemaTypeUtil.INTEGER_TYPE;
 import static org.openapitools.codegen.CodegenConstants.API_PACKAGE;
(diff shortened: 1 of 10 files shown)


altro3:
@graemerocher review this also, please

graemerocher (approved):

Issue #2648, opened 2026-03-31
Duplicate dependency flexmark-html2md-converter


auke-:
### Expected Behavior

Maven build without warnings.

### Actual Behaviour

When using Micronaut 4.10.x with the OpenAPI module (through micronaut-platform parent) Maven emits a warning about a duplicate dependency:

```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for foo:bar:jar:0.1
[WARNING] 'dependencyManagement.dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: com.vladsch.flexmark:flexmark-html2md-converter:jar -> version ${html2md.converter.version} vs ${flexmark.version} @ io.micronaut.platform:micronaut-platform:4.10.7, <HOME>/.m2/repository/io/micronaut/platform/micronaut-platform/4.10.7/micronaut-platform-4.10.7.pom, line 3131, column 19
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

This is probably caused by the the duplicate `flexmark-html2md-converter` entry in https://github.com/micronaut-projects/micronaut-openapi/blob/7.0.x/gradle/libs.versions.toml#L64-L65.

### Steps To Reproduce

Create a new Micronaut 4.10.x project using Maven as the build tool and run `./mvnw clean`

### Version

4.10.x

See https://github.com/micronaut-projects/micronaut-platform/issues/2410

auke-:
@sdelamo would it be possible to get this fixed in the [4.10.x branch](https://github.com/micronaut-projects/micronaut-platform/tree/4.10.x) of [micronaut-platform](https://github.com/micronaut-projects/micronaut-platform/) (see [#2410](https://github.com/micronaut-projects/micronaut-platform/issues/2410) and [PR #2549](https://github.com/micronaut-projects/micronaut-openapi/pull/2649))? 🙏

auke-:
@altro3 could you take a look at [PR #2649](https://github.com/micronaut-projects/micronaut-openapi/pull/2649), as i think you have added the relevant lines in https://github.com/micronaut-projects/micronaut-openapi/commit/9adebf6583515be7d8223e33add16c68bf98670a 🙏  The warning is quite annoying, especially in agentic workflows.

PR #2649, opened 2026-03-31
fix: remove duplicate flexmark-html2md-converter dependency (#2648)

auke-:
The duplicate flexmark-html2md-converter dependency is causing Maven to emit warnings since dependencies must be unique.

Since the property managed-html2md-converter isn't actually used and is replaced by managed-flexmark we can safely remove it and its corresponding dependency declaration.

See #2648

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,8 +2,6 @@
 managed-swagger = "2.2.50"
 managed-javadoc-parser = "0.3.1"
 managed-jsoup = "1.22.2"
-# deprecated, use managed-flexmark
-managed-html2md-converter = "0.64.8"
 managed-flexmark = "0.64.8"
 managed-parboiled = "1.4.1"
 managed-pegdown = "1.6.0"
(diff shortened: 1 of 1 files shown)


sdelamo (changes requested):
this should target 7.0.x as removing a managed dependency will be a breaking change in the bom.

auke-:
> this should target 7.0.x as removing a managed dependency will be a breaking change in the bom.

I don't think it is removing a dependency, it is removing a duplicate.

auke-:
I've rebased the change on the [7.0.x](https://github.com/micronaut-projects/micronaut-openapi/tree/7.0.x) branch.

I could be wrong, but i don't think this is a breaking change as it only removes a duplicate dependency. The [managed-flexmark](https://github.com/micronaut-projects/micronaut-openapi/blob/7.0.x/gradle/libs.versions.toml#L7) property and [flexmark dependencies](https://github.com/micronaut-projects/micronaut-openapi/blob/7.0.x/gradle/libs.versions.toml#L65-L66) are still there. The [CI job](https://github.com/auke-/micronaut-openapi/actions/runs/26808847175/job/79033207087#step:9:1941) fails on the missing `html2md-converter` dependency, which is the duplicate of `flexmark-html2md-converter`.

PR #2656, opened 2026-04-04
Fixed container with HttpReponse types interpretation

altro3:


openapi/src/main/java/io/micronaut/openapi/visitor/ElementUtils.java
```diff
@@ -88,37 +89,55 @@ public final class ElementUtils {
...
-    public static final List<String> CONTAINER_TYPES = List.of(
+    public static final List<String> ARRAY_CONTAINER_TYPES = List.of(
+        "io.reactivex.Flowable",
+        "io.reactivex.Observable",
+        "io.reactivex.rxjava3.core.Flowable",
+        "io.reactivex.rxjava3.core.Observable",
+        "reactor.core.publisher.Flux",
+        "kotlinx.coroutines.flow.Flow",
+        "org.reactivestreams.Publisher",
+        "io.smallrye.mutiny.Multi"
```
graemerocher, line 108: this duplicates logic in https://github.com/micronaut-projects/micronaut-core/blob/5.0.x/core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java probably better to use that

graemerocher (changes requested):

graemerocher (approved):

PR #2730, opened 2026-05-26
Add warning about config with KSP

mmwinther:
- **Add warning about application.yml with ksp**
- **Add example for config with KSP**


lissam1:
It will be supported after this pr https://github.com/micronaut-projects/micronaut-openapi/pull/2745

PR #2745, opened 2026-06-12
Restore application.yml config for OpenAPI processors

### LINKED ISSUE #2729: Config in application.yml has no effect with ksp
### Expected Behavior

[The documentation](https://micronaut-projects.github.io/micronaut-openapi/6.20.0/guide/index.html#applicationConfiguration) refers to `application.yml` as being a possible configuration source for Micronaut OpenAPI processing.

Unfortunately this has no effect when using KSP. A warning should be added to the documentation about it.

There should also be an example of the most elegant configuration when using KSP e.g.

```
ksp {
    arg("micronaut.openapi.project.dir", projectDir.toString())
    arg("micronaut.openapi.property.naming.strategy", "SNAKE_CASE")
}
```

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

5.0.0

lissam1:
## Summary
- Add SnakeYAML as a transitive dependency of micronaut-openapi so annotation processing can load application.yml without user-side ksp/kapt SnakeYAML dependencies.
- Cover KSP and KAPT application.yml configuration with a naming-strategy regression test.
- Remove the explicit KSP SnakeYAML dependency from the Spring KSP test suite to prove micronaut-openapi supplies YAML loading.

Closes #2729

### DIFF
--- a/openapi/build.gradle.kts
+++ b/openapi/build.gradle.kts
@@ -12,6 +12,7 @@ dependencies {
     implementation(mn.micronaut.core)
     implementation(mn.micronaut.inject)
     implementation(mn.micronaut.http)
+    implementation(mn.snakeyaml)
 
     api(projects.micronautOpenapiAnnotations)
     api(projects.micronautOpenapiCommon)
--- a/openapi/build.gradle.kts
+++ b/openapi/build.gradle.kts
@@ -12,7 +12,7 @@ dependencies {
     implementation(mn.micronaut.core)
     implementation(mn.micronaut.inject)
     implementation(mn.micronaut.http)
-    implementation(mn.snakeyaml)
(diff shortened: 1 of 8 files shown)


graemerocher:
this seems like a documentation issue, why do we need to depend on SnakeYAML?

lissam1:
> this seems like a documentation issue, why do we need to depend on SnakeYAML?

OpenAPI reads application.yml during annotation processing, so KSP/KAPT need SnakeYAML on the isolated processor runtime classpath; runtimeOnly(mn.snakeyaml) restores application.yml support
  without making it a compile dependency.
if you don't want a dependency in snakeyaml we can document that users need to add it to use application.yml

PR #2751, opened 2026-06-17
feat(openapi): opt-in $dynamicAnchor / $dynamicRef emission for recursive types and single-variable generics

### LINKED ISSUE #2748: feat: emit $dynamicAnchor / $dynamicRef from Java generics and recursive types (opt-in)
### Feature description

**Proposal:** add an opt-in mode that auto-emits `$dynamicAnchor` / `$dynamicRef` (JSON Schema 2020-12, [spec §7.7](https://json-schema.org/draft/2020-12/json-schema-core#section-7.7)) from Java types — both recursive inheritance trees and generic/template types — so a `Page<T>` / `ApiEnvelope<T>` produces one reusable template schema plus inline bindings at the point of use, instead of a duplicated concrete schema per item type.

If landed, Micronaut OpenAPI would be the **first spec producer to auto-generate `$dynamicAnchor`/`$dynamicRef` from language types**. A public compatibility lab tracks this across the ecosystem — https://github.com/aqeelat/openapi-dynamicref-adoption-tracker — and no producer currently emits these keywords from source; Pydantic, FastAPI, tsoa, springdoc, and others all hit the same wall. The reader side (consuming these keywords in a generator) is proven — Orval merged support in May (https://github.com/orval-labs/orval/pull/3353) — but no producer writes them yet.

### Why Micronaut is well positioned

1. **The authoring layer already exists.** `@Schema` annotation properties can already set `$dynamicRef` / `$dynamicAnchor` / `$anchor` / `$id` when targeting OpenAPI 3.1, and the keywords survive schema merges. What's missing is *auto-emission* — today they're authorable but nothing generates them from types.
2. **Java generics are visible at the processing stage.** Because Micronaut is compile-time AOT, the annotation processor sees full generic type info — type arguments are not erased. Today the visitor resolves each type variable to a concrete type at every binding site, producing duplicate concrete schemas. The change is to preserve the type variable as a `$dynamicAnchor` and emit `$dynamicRef` consumers when an opt-in mode is on.

### User-impact evidence

[#1288](https://github.com/micronaut-projects/micronaut-openapi/issues/1288) (closed, 2023) reports that `Iterable<T>` / `T[]` fields lose the `T` type, and the generated schema materializes `ListElement<String>` as a duplicated `ListElement_String_` concrete schema. That's exactly the concrete-materialization pain this feature resolves — preserving the generic shape instead.

### What the generated spec should reproduce

The realistic target is the tracker's combined showcase — https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/petstore-dynamicref-showcase.yaml — which exercises all patterns in a Petstore-style API. Java signatures map to emissions as follows:

| Java signature | Emission |
|---|---|
| `@Get ApiEnvelope<Pet> getPet()` | inline binding at the response schema |
| `@Get ApiEnvelope<Page<Pet>> listPets()` | named `Page<Pet>` binding + inline envelope binding (composes) |
| `class BaseCategory { List<BaseCategory> children }` + `Localized extends BaseCategory` | recursive `$dynamicAnchor` |
| `class Folder<F extends Folder<F,R>, R>` | multi-anchor template (two `$dynamicAnchor`s) |

Key point on binding location: a parameterized type used directly at an endpoint (`ApiEnvelope<Pet>`) emits the binding **inline at the response**, because there is no named `ApiEnvelopePet` Java type — synthesizing one (today's behavior) invents a type that doesn't exist in the source. A named binding in `components/schemas` is emitted only when the source declares a named subtype (`class PaginatedUserResponse extends Page<User>`).

### Implementation plan

Opt-in flag (default unchanged): `micronaut.openapi.generic-schema-mode: dynamic-ref` (default `concrete` = today's behavior). Zero back-compat risk.

Two independent sub-modes, recursion first since it needs no type-variable preservation:

**Sub-mode A — recursive-anchor emission.** For self-referential types through inheritance, detect that a property's type equals (or recurses into) the enclosing schema, then emit `$dynamicAnchor` at the base and extending schema and `$dynamicRef` at the recursive property instead of a self-`$ref`. Target output (recursive category tree, OAS 3.1):

```yaml
BaseCategory:
  $dynamicAnchor: category
  properties:
    children:
      type: array
      items:
        $dynamicRef: '#category'
LocalizedCategory:
  $dynamicAnchor: category
  allOf:
    - $ref: '#/components/schemas/BaseCategory'
    - type: object
      properties:
        displayName: { type: string }
```

**Sub-mode B — generic-template preservation.** For generic types (`Page<T>`, `ApiEnvelope<T>`), preserve the type variable instead of resolving it:

1. **Template emission** — when a generic type `Foo<T>` is encountered, treat an unresolved type variable as a generic slot, emit a `$dynamicAnchor` named after it (sanitized), and at each use of `T` (e.g. `List<T> items`) emit `$dynamicRef: '#<anchor>'` instead of resolving to a concrete schema. Multi-parameter types (`Map<K,V>`, `Folder<F,R>`) emit one anchor per type variable.
2. **Binding emission** — at a parameterized site (`Foo<User>`), emit the binding inline at that site (`$ref` to the template + a sibling `$defs.<anchor>` with `$dynamicAnchor` + `$ref` to the concrete type); emit a named binding in `components/schemas` only for declared named subtypes.
3. **Cycle safety** — scope tracking so recursive generic templates (`class Tree<T> { List<Tree<T>> children; }`) resolve `$dynamicRef` to the enclosing template name, terminating at a named reference rather than re-inlining.

Target output — single-wrap (`@Get ApiEnvelope<Pet>`), inline at the response:

```yaml
paths:
  /pets/{petId}:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $defs:
                  dataType:
                    $dynamicAnchor: dataType
                    $ref: '#/components/schemas/Pet'
                $ref: '#/components/schemas/ApiEnvelopeTemplate'
components:
  schemas:
    ApiEnvelopeTemplate:
      $defs:
        dataType:
          $dynamicAnchor: dataType
          not: {}
      properties:
        data:
          $dynamicRef: '#dataType'
```

Target output — double-wrap (`@Get ApiEnvelope<Page<Pet>>`), proving bindings compose:

```yaml
paths:
  /pets:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $defs:
                  dataType:
                    $dynamicAnchor: dataType
                    $ref: '#/components/schemas/PaginatedPetItems'   # itself a Page<Pet> binding
                $ref: '#/components/schemas/ApiEnvelopeTemplate'
components:
  schemas:
    PaginatedPetItems:
      $defs:
        itemType:
          $dynamicAnchor: itemType
          $ref: '#/components/schemas/Pet'
      $ref: '#/components/schemas/PaginatedTemplate'
```

Compared to today's duplicated `PaginatedPetItems` / `ApiEnvelopePet` / etc. concrete schemas.

**Planned PRs** (will be linked here once opened):

- [x] PR 1: opt-in flag (`generic-schema-mode`) + recursive-anchor emission (Sub-mode A) + tests
- [x] PR 2: generic-template preservation (Sub-mode B) + binding emission + cycle safety + tests

### Open questions

- **Anchor naming:** keep the Java type-variable name as the anchor (e.g. `itemType`), or map to conventional `T`/`E`? Keeping the variable name is more faithful and avoids collisions.
- **`allOf` flattening:** ensure emitted templates/bindings survive Micronaut's inheritance-flattening passes.

### Offer to implement

Happy to contribute, scoped as the two PRs above — flag + recursion first, then generics. I can combine into a single PR if reviewers prefer one landing.

### Compatibility evidence & fixtures

Cross-generator compatibility matrix and validator-backed fixtures: https://github.com/aqeelat/openapi-dynamicref-adoption-tracker

Target shapes the generated spec should match:
- Showcase (realistic API, all patterns): [`petstore-dynamicref-showcase.yaml`](https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/petstore-dynamicref-showcase.yaml)
- Recursive: [`fixtures/recursive-category-tree.yaml`](https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/recursive-category-tree.yaml)
- Inline binding: [`fixtures/paginated-response.yaml`](https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/paginated-response.yaml), [`fixtures/api-envelope.yaml`](https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/api-envelope.yaml)
- Named binding (declared subtypes): [`fixtures/generic-schema-binding.yaml`](https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/generic-schema-binding.yaml)

---
*This issue was drafted with assistance from AI tooling. The submitter has reviewed and validated the contents.*

aqeelat:
## Summary

Adds an opt-in mode (`micronaut.openapi.schema.dynamic-refs.enabled`, OpenAPI 3.1 only, default off) that emits JSON Schema 2020-12 `$dynamicAnchor` / `$dynamicRef` from Java types, so consumers that understand dynamic scope (validators, SDK generators) resolve recursive and generic references against the active type rather than a static base.

Default behavior is unchanged — the feature is gated behind the flag and on `isOpenapi31()`.

Closes #2748.

## What it does

**Recursive types.** A self-referential property emits `$dynamicRef`, and the recursive base schema plus every subtype declare a matching `$dynamicAnchor`. Dynamic-scope resolution then switches the reference to the active subtype at the point of use (e.g. `LocalizedCategory.children` resolves to `LocalizedCategory`, not `BaseCategory`).

**Single-variable generics.** A generic wrapper such as `Response<T>` / `Page<t>` collapses from one duplicated concrete schema per binding (`Response_Pet_`, `Response_Group_`, …) into one reusable template plus an inline `$defs` binding at each usage site. The template carries a `$dynamicAnchor` (role-derived: `itemType` for collection/array elements, `dataType` otherwise) and a `$defs` placeholder; each binding re-declares the anchor pointing at the concrete type. The wrapper instance validates against the template only — the concrete schema applies solely where the `$dynamicRef` sits. The `allOf`-binding alternative is intentionally **not** used: it would apply the concrete schema to the whole wrapper instance under `allOf` semantics and reject valid parameterized instances.

Example output for `Response<Pet>`:

```yaml
components:
  schemas:
    Response:                  # one template
      \$dynamicAnchor: dataType
      properties:
        data: { \$dynamicRef: '#dataType' }
      \$defs:
        dataType: { \$dynamicAnchor: dataType, not: {} }
paths:
  /pet:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:                                          # inline binding
                \$ref: '#/components/schemas/Response'
                \$defs:
                  dataType: { \$dynamicAnchor: dataType, \$ref: '#/components/schemas/Pet' }
```

## Implementation notes

- All changes are internal (`ConfigUtils`, `OpenApiConfigProperty`, `SchemaDefinitionUtils` are `@Internal`); no public API surface, no `@since` tags required.
- **`$defs` workaround (temporary, documented in code + guide):** swagger-core 2.2.x `Schema` has no `$defs` field and `addExtension` rejects non-`x-` keys, so `$defs` is attached via reflection on `Schema`'s `extensions` map (which the serializer emits verbatim through `@JsonAnyGetter`). This is isolated in one helper (`setSchemaDefs`) and is structured to auto-switch to native `set$defs` once swagger-core ships it — tracked upstream in swagger-api/swagger-core#4469.

## Scope / limitations (documented in `dynamicRefs.adoc`)

- Single type variable only; multi-parameter generics (`Map<K,V>`) keep concrete schemas.
- Applies where the parameterized type is used directly (return type / field) with a reference-type argument. Primitives, arrays, collections, maps, `@Schema`-annotated generics, named subtypes (`class X extends Foo<Bar>`), nested parameterized bindings, and self-referential generics keep their default concrete schemas.
- Recursive subtypes that introduce a *different* recursive property of their own fall back to a normal `$ref`.

## Testing

- New: `OpenApi31DynamicAnchorsSpec` (recursion: 7 tests), `OpenApi31DynamicGenericsSpec` (generics: 8 tests) covering template + binding shape, role-derived anchors, back-compat (flag off), OpenAPI 3.0 fallback, primitive-arg fallback, raw-before-parameterized ordering, self-referential generics, and `$defs` JSON serialization.
- Regression: `OpenApiSchemaGenericsSpec` (incl. recursive/nested generics), `OpenApi31Spec`, `OpenApiPageSpec` pass unchanged.
- `./gradlew cM spotlessCheck` clean.

---
*This PR was drafted with assistance from AI tooling. The submitter has reviewed and validated the contents.*


openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
```diff
@@ -293,11 +323,289 @@ private SchemaDefinitionUtils() {
...
+        return "dataType";
+    }
+
+    /**
+     * Whether the given parameterized type should be emitted as a dynamic-ref generic binding:
+     * dynamic-refs mode is on, it has exactly one type argument, it is not a collection / map,
+     * and the argument is a concrete reference type (not an unresolved type variable, primitive,
+     * array, collection or map — those cannot form a {@code $ref} binding slot).
+     */
+    private static boolean isGenericBindingCandidate(ClassElement type, Map<String, ClassElement> typeArgs) {
```
altro3: Need to move helper methods to other classes. This class is so fat already

aqeelat: Done — extracted them into a new `io.micronaut.openapi.visitor.DynamicRefUtils` (add2c3ab). `SchemaDefinitionUtils` shrinks by ~310 net lines; the few call sites now delegate to `DynamicRefUtils.*`, and `populateSchemaProperties` widens to package-private so the new class can drive the generic template build. No behavior change.

altro3 (approved):

aqeelat:
Hi @graemerocher, can you please review the PR? @altro3 already approved it.

graemerocher:
Thanks for the contribution!

PR #2779, opened 2026-07-24
Use generated files for default OpenAPI resources

graemerocher:
## Summary

- Generate default OpenAPI specifications, views, themes, and AsciiDoc through VisitorContext.visitMetaInfFile.
- Track contributing application, endpoint, group, version, and extra-schema elements for incremental compilation.
- Preserve explicit filesystem destinations and AsciiDoc append behavior.
- Add classloader-based coverage for default specifications, views, multiple documents, nested resources, and AsciiDoc output.

## Motivation

The current implementation writes files to the classes directory without javac being aware of those outputs. This breaks incremental compilation and also means the processor cannot be used in in-memory scenarios where there is no disk access.

## Verification

- ./gradlew :micronaut-openapi:test — 504 tests passed, 3 skipped
- ./gradlew :micronaut-openapi:spotlessCheck :micronaut-openapi:check
- ./gradlew check docs
- git diff --check

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/view/OpenApiViewConfig.java
+++ b/openapi/src/main/java/io/micronaut/openapi/view/OpenApiViewConfig.java
@@ -21,6 +21,7 @@ import io.micronaut.core.io.scan.DefaultClassPathResourceLoader;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.inject.visitor.VisitorContext;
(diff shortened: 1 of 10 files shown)


sdelamo (approved):

graemerocher:
Addressed Copilot feedback in commit 829e50017. Default generated OpenAPI specifications now register swagger/<filename> and swagger with VisitorContext.addGeneratedResource; default views register swagger/views and swagger; default AsciiDoc registers swagger/<filename> and swagger. visitMetaInfFile remains the actual generated-file writer.

PR #2782, opened 2026-07-30
Fix OpenAPI incremental processing inputs

graemerocher:
## Summary

- declare the annotations supported by aggregating OpenAPI visitors instead of inheriting the wildcard
- record only endpoint and group elements that actually contribute to generated OpenAPI output
- add regression coverage for supported annotations and unrelated originating elements

## Why

Wildcard matching and eagerly recorded origins cause incremental compilers to treat every source as an input to the shared OpenAPI document. This prevents unrelated source changes from compiling incrementally.


andriy-dmytruk (approved):

graemerocher:
Regression coverage has been updated with the broader package patterns suggested in review:

- OpenApiApplicationVisitor: io.swagger.v3.oas.annotations.* plus the Micronaut group annotations
- OpenApiControllerVisitor: io.micronaut.http.annotation.* and io.swagger.v3.oas.annotations.*

This avoids omissions when new annotations are added while still eliminating the global wildcard that made every source an aggregating candidate. The tests also verify that unrelated controller/group inputs are not recorded as originating elements. OpenApiApplicationVisitorSpec passes all 15 tests and Spotless is green.

PR #2793, opened 2026-08-11
  Fix duplicate schema property for @JsonProperty getters

### LINKED ISSUE #2791: Explicit @Schema(name=...) on a class causes the last bean property to be duplicated under the schema's own name
### Expected Behavior

A class with an explicit @Schema(name="CustomName") override should generate a schema named CustomName containing only its real properties.

### Actual Behaviour

The LAST declared bean property is duplicated a second time in the properties map, under a bogus key equal to the class's own @Schema name value, and its description is overwritten with the class-level @Schema description instead of the original field's description.

### Steps To Reproduce

Minimal reproduction:

``` java
​import io.swagger.v3.oas.annotations.media.Schema;
import java.util.Set;

@Schema(name = "CustomName", description = "My class description")
public class Foo {
    @Schema(description = "the from set")
    private Set<String> from;

    public Set<String> getFrom() { return from; }
    public void setFrom(Set<String> from) { this.from = from; }
}
```
​

Expose Foo from a controller method and generate the OpenAPI spec.

Generated (buggy) output:

``` yaml
components:
  schemas:
    CustomName:
      type: object
      properties:
        from:
          type: array
          description: the from set
          items:
            type: string
        CustomName:              # <-- bogus duplicate property, keyed by the schema's own name
          type: array
          description: My class description   # <-- class-level description leaked in
          items:
            type: string
      description: My class description
```
​
Removing the name= attribute (letting the schema default to the simple class name Foo) makes the bogus property disappear.

Also reproduced with a custom short name value (not just a fully-qualified class name), so it's tied to the mere presence of an explicit name override, not its content.

This was working correctly in micronaut-openapi 6.19.3 (last version usable with Micronaut 4). 6.19.3 is binary-incompatible with Micronaut 5 (AnnProcessorEnvironment extends DefaultEnvironment, which became final in Micronaut 5), so downgrading to avoid this bug is not viable for Micronaut 5 users.

### Environment Information

- micronaut-openapi 7.0.0, 7.1.0, 7.1.2 — reproduced identically on all three
- Micronaut 5
- Windows, Gradle 9.6.1, JDK 25

### Version

7.0.0 / 7.1.0 / 7.1.2

lissam1:
issue : 
 When a Java bean getter is annotated with `@JsonProperty`, the OpenAPI visitor processed it twice:
  1. As a normal bean property.
  2. In the additional `@JsonProperty`-method pass.
  For a class with an explicit `@Schema(name = "...")`, the second pass could inherit the class-level schema name and add a duplicate property keyed by that
  schema name.
  The additional Jackson-method pass now skips methods that are already the read method of a discovered bean property. Standalone non-bean methods annotated with
  `@JsonProperty` continue to be processed.
Published fix locally and test it against reproducer now generates only the 'from' property.
Closes : #2791

### DIFF
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
@@ -1949,6 +1949,9 @@ public final class SchemaDefinitionUtils {
                 if (!methodEl.hasAnnotation(JsonProperty.class) || isVoid(methodEl.getReturnType())) {
                     continue;
                 }
+                if (beanProperties.stream().anyMatch(property -> property.getReadMethod().filter(methodEl::equals).isPresent())) {
+                    continue;
+                }
                 var methodJavadoc = Utils.getJavadocParser().parse(methodEl.getDocumentation().orElse(null));
                 var methodReturnType = methodEl.getReturnType();
 
--- a/openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
+++ b/openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
@@ -1950,10 +1950,9 @@ public final class SchemaDefinitionUtils {
 
             // also need to check methods with @JsonProperty annotation
             for (var methodEl : classEl.getMethods()) {
-                if (!methodEl.hasAnnotation(JsonProperty.class) || isVoid(methodEl.getReturnType())) {
-                    continue;
-                }
-                if (beanPropertyReadMethods.contains(methodEl)) {
+                if (!methodEl.hasAnnotation(JsonProperty.class)
+                    || isVoid(methodEl.getReturnType())
+                    || beanPropertyReadMethods.contains(methodEl)) {
                     continue;
                 }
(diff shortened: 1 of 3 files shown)


openapi/src/main/java/io/micronaut/openapi/visitor/SchemaDefinitionUtils.java
```diff
@@ -1949,6 +1949,9 @@ static void populateSchemaProperties(OpenAPI openApi, VisitorContext context, El
                 if (!methodEl.hasAnnotation(JsonProperty.class) || isVoid(methodEl.getReturnType())) {
                     continue;
                 }
+                if (beanProperties.stream().anyMatch(property -> property.getReadMethod().filter(methodEl::equals).isPresent())) {
```
graemerocher: is there a loop somewhere else that handles these beanProperties that can be reused instead of streaming them all again?

lissam1: we have  `beanProperties` inside` processPropertyElements(...)` but seems dosn't expose methods processed.
i did this https://github.com/micronaut-projects/micronaut-openapi/pull/2793/commits/623d98269764b5fb8f273d7d241952b1d6caf9f4 it will streams `beanProperties` once, then reuse it to check jackson methods.

altro3 (approved):

PR #2830, opened 2026-09-16
Multi-language documentation snippets (Java, Kotlin, Groovy, Python)

graemerocher:
Backs every application-code sample of the user guide with compiled and tested sources in Java, Kotlin, Groovy **and Python**, and renders the `dependency:` macros for the Python compiler.

Part of the multi-language documentation effort built on micronaut-projects/micronaut-build#939.

Requires micronaut-build **8.1.2** (`io.micronaut.build.internal.python` plugin with `compilerArgs`, Python `dependency:` rendering) and micronaut-core **5.2.3** (the version catalog is bumped to both).

## What was ported

* `docs-examples/example-python` next to `example-java`, `example-kotlin` and `example-groovy` (the projects the `snippet::[project-base="docs-examples/example"]` macros resolve), plus `example-config-*`, `example-groups-*` and `example-versions-*` in all four languages for the samples that need their own `@OpenAPIDefinition`, `openapi.properties` or `application.yml` (property placeholders, tag generation, `@OpenAPIGroup`/`@OpenAPIGroupInfo`, `@Version`).
* 27 inline `[source,java]` blocks converted to `snippet::` macros (31 snippet blocks per language): `@OpenAPIDefinition`, controllers, swagger annotations, schemas (POJOs, meta annotations, generics, naming, annotation resolution), `@AccessorsStyle`, `@OpenAPIDecorator`, `@OpenAPIExclude`, `@OpenAPIExtraSchema`, `@OpenAPIGroup`, `@OpenAPIGroupInfo`, `@OpenAPIInclude`, OAuth 2 `@SecurityScheme` / `@SecurityRequirement`, server filters, custom serializers, dynamic refs, tags generation, versions, processing options.
* Every example project has tests asserting the generated `META-INF/swagger/*.yml` (JUnit 5 for Java and Kotlin, Spock for Groovy, the Python compiler `@MicronautTest` for Python): 16 projects, 80 tests.
* `openapi-example-{base,java,kotlin,groovy,python}` convention plugins in `buildSrc`; the docs Kotlin example now uses KSP (kapt silently runs no processor in this build, also for the untouched `example-kotlin-kapt`).
* Fixes on the way: the Groovy swagger annotations example had broken tag markers, the Kotlin one lacked `@Parameter`/`@NotBlank`, the two Spring `snippet::` macros pointed at non-existent projects (`test-suite-java-spring`, `test-suite`), the `@OpenAPIDefinition` "generated YAML" and the versions file names/titles did not match the examples, `application.yml` placeholders need snakeyaml on the annotation processor path (now documented with a `dependency:` macro).
* `.github/workflows/python.yml` (copied from the project template) runs `./gradlew pythonCheck -Ppython-ci`.
* Verified against core 5.2.3 / micronaut-build 8.1.2: no test is disabled and no source workaround is left, so there is no `DISABLED_TESTS.md`.

## Not converted

* `micronautOpenApiAnnotations/openApiManagement.adoc` and `openApiSecurity.adoc`: the `[source,java]` blocks show what `@OpenAPIManagement` / `@OpenAPISecurity` are *mapped to* (a bare annotation expression), not application code.
* The `[source,kotlin]` / `[source,groovy]` blocks of the guide are all Gradle build snippets (KAPT/KSP arguments, Spring tasks, views system properties) and stay as they are.
* The Spring `MyController` snippets are Java-only (`language="java"`, `project="test-suite-spring-java"`).

## Python notes

* No test is `@Disabled` and there is no `java.type(...)` call or source workaround left: the swagger annotations and models are imported with their Java names (`from io.swagger.v3.oas.annotations import Operation`, `from io.swagger.v3.oas.models import OpenAPI`, nested enums as `from io.swagger.v3.oas.models.PathItem import HttpMethod`), the generic models (`Response[T]`, `Page[T]`, `MyJaxbElement[T]`) are introspected generic dataclasses, `LocalizedCategory` extends the `BaseCategory` dataclass, `Pet.type` is an optional enum and `MyDto` has `JAXBElement[...]` attributes, like the Java examples. The main and test Python sources are compiled as separate source sets like the other languages.
* Every example project passes its project directory to the OpenAPI visitor explicitly (`-Amicronaut.openapi.project.dir` for Java and Python, `-D` for Groovy, the KSP `arg` as before), like `test-suite-server-generator-java` does: the visitor otherwise derives it from the compiler's working directory or a `micronaut.openapi.project.dir` system property, both of which belong to whichever project first started a shared Gradle worker daemon (the Micronaut KSP processor copies its options to system properties), so `openapi.properties` / `application.yml` were only found depending on the task order (`example-config-java` failed after `example-kotlin` in the same build). The Python compiler cannot derive the directory from its generated files at all, so the option is required there.
* Nested classes are not supported in Python, so `UserDto` is a top-level class in every language; `@AccessorsStyle` has no Python equivalent (attributes are exposed directly), the Python `Person` is an introspected dataclass.
* KSP and the Groovy AST transformation cannot enumerate the classes of a *source* package (`@OpenAPIExtraSchema(packages = ...)`), Groovy doc comments are not available to the visitor, and Kotlin non-null properties are `required`: the Kotlin/Groovy tests document these differences with comments.

## Verification

```
./gradlew :micronaut-docs-examples:micronaut-example-{java,kotlin,groovy}:check            # 14 tests each
./gradlew :micronaut-docs-examples:micronaut-example-{config,groups,versions}-{java,kotlin,groovy}:check   # 2 tests each
./gradlew pythonCheck -Ppython-ci   # 20 Python tests (example-python 14, config 2, groups 2, versions 2), GraalVM 25
./gradlew docs                      # build/docs/guide/index.html: 31 data-lang="python" blocks, 35 java, 36 kotlin, 38 groovy, 7 pyproject.toml dependency tabs
```


dstepanov (commented):
### Review: multi-language docs port (pyronaut#139)

Verdict: ⚠️ minor issues
Findings:
- [severity: medium] docs-examples/example-python/src/main/python/micronaut/configuration/openapi/docs/schemas/MyAnn.py:7-9 and PetController.py:18. The Python version of the "meta annotation" is a plain function decorated with `@Schema(...)`, attached as `Annotated[Pet, Body, MyAnn]`. It reads oddly next to the Java `@interface` and is not shown to work. `SchemasTest.test_meta_annotation_applies_the_schema` only asserts `$ref == #/components/schemas/MyPet`, but `Pet` itself already carries `@Schema(name="MyPet")` (schemas/Pet.py:14), so the test passes even if `MyAnn` is ignored. The same weakness exists in the Java, Kotlin and Groovy tests. Fix it by giving the meta annotation a value that differs from the type-level `@Schema`, or by asserting on a parameter or field that has no type-level schema, so the test proves the meta annotation is applied in each language.
- [severity: low] docs-examples/example-python/src/main/python/micronaut/configuration/openapi/docs/Application.py:50-51. The Python `@OpenAPIExclude(packages=…)` uses `micronaut.configuration.openapi.docs.exclude.package1/package2`, while Java, Kotlin and Groovy show `io.exclude.package.with.controllers{,2}`. The `exclude` tag is rendered in the guide (openApiExclude.adoc), so the Python tab shows different values from the other tabs.
- [severity: low] src/main/docs/guide/configuration/applicationConfiguration.adoc:18-25. The `application.yml` placeholder example dropped from three properties (`my.api.version`, `my.api.title`, `my.api.api-description`) to only `my.api.title`. The page now sends readers to openapi.properties for the rest and reuses the same snippet as propertiesFileConfiguration.adoc. The old block did have title and version swapped, but the page no longer shows several `application.yml` placeholders together, so both configuration pages now show one example.
- [severity: low] buildSrc/src/main/kotlin/io.micronaut.build.internal.openapi-example-java.gradle.kts:11 and …-kotlin.gradle.kts:12. These add `micronaut-validation`, the runtime library, to `annotationProcessor`/`ksp` instead of `micronaut-validation-processor`. It has no effect, and it carries over the old `kapt(mnValidation.micronaut.validation)` mistake. Use the processor artifact or drop the line.
- [severity: low] The PR body says kapt "silently runs no processor in this build", which also affects the untouched `docs-examples/example-kotlin-kapt`. That project has no tests, so the breakage stays silent. It is outside this PR's scope, but worth a follow-up issue.
- Checked and OK. There are no production (`*/src/main/**` library) changes and no deleted files. All ~790 deleted lines are either inline `[source,java]` blocks, each replaced by a `snippet::` macro whose Java source matches the removed code (same annotations, values and callouts), or boilerplate moved from the three `docs-examples/example-*/build.gradle.kts` files into the new `openapi-example-*` convention plugins (repositories, `-parameters`, test logging, `skipDocumentation`). Explanatory prose was kept. Where it changed, it was to correct errors: `Response<Pet>` → `Response_Pet_`; swagger-1/2.yml → service-1.0.0-{1,2}.yml with the matching title and version; `openapi: 3.0.3` → `3.0.1`, which is the swagger-models default; the JAXBElement mapping keys now use `jakarta.xml.bind.JAXBElement` and real package names. All `snippet::` macros, including the comma-separated multi-class ones, resolve to existing files with the referenced tags in all four languages. The Spring macros now point to the existing `test-suite-spring-java` project with `language="java"`. Callouts match across languages. There are no `java.type`, try/except import hacks, `@Disabled` tests or `DISABLED_TESTS.md`; `io.swagger.*` is imported directly (A2 and A16 are fixed in 5.2.3). The Python tests make real assertions on the generated YAML. Versions are micronaut-build 8.1.2 and core 5.2.3. New settings includes and the `jakarta-xml-bind-api` catalog entry are consistent.
- CI: audit passed. build and Python detect were still pending at review time.

dstepanov (approved):
Approving — the findings in my previous review are minor.

